PR #17, opened 2026-04-01
auth: add file-based credential storage (PRINFRA-123)

somanshreddy:
### What This Adds
This PR adds **persistent local authentication** to the CLI.

`heygen auth login` reads an API key from stdin (interactive prompt or pipe), stores it in the CLI config directory with **restrictive file permissions** (0600), and returns structured JSON. This gives the CLI a real local login flow instead of requiring users to export `HEYGEN_API_KEY` for every session.

### Login Flow
`auth login` **skips the global pre-run auth check** via Cobra annotations (`skipAuth: "true"`). Before this change, `PersistentPreRunE` tried to resolve credentials for every command, which blocked the login command itself when no API key existed yet.

### Credential Resolution Order
Credential lookup uses a chain resolver with priority order:

1. `HEYGEN_API_KEY` env var — for CI, Docker, and agent workflows
2. `~/.heygen/credentials` file — written by `heygen auth login`

### Error Handling
The auth layer distinguishes between:
- A credential source that is **absent** (`ErrNotConfigured`) — falls through to the next resolver
- A credential source that is **present but broken** (permission denied, empty file) — surfaces immediately with an actionable hint pointing to the actual credential path

### No `--key` flag
Industry standard (gh, AWS, Stripe, gcloud) is to never pass secrets as CLI flags — they leak into `ps` output and shell history. Instead:
- **Agents/CI**: `HEYGEN_API_KEY` env var (no login needed)
- **Humans (interactive)**: `heygen auth login` → prompt with hidden input
- **Humans (scripted)**: `echo "$KEY" | heygen auth login`

### Example Outputs

**Successful login:**
```
$ echo "$KEY" | heygen auth login
{
  "message": "API key saved to ~/.heygen/credentials"
}
```

**Empty input:**
```
$ echo "" | heygen auth login
{
  "error": {
    "code": "usage_error",
    "message": "no API key provided"
  }
}
```

**No credentials configured (on any API command):**
```
$ heygen video list
{
  "error": {
    "code": "auth_error",
    "message": "no API key found",
    "hint": "Set HEYGEN_API_KEY env var or run: heygen auth login"
  }
}
```

**Linear issue:** PRINFRA-123

### Automated (16 tests)
- **Chain resolver** (4): first-wins, fallthrough on ErrNotConfigured, all-not-configured → auth error, broken-source → surfaces immediately
- **File resolver** (4): reads key, missing file → ErrNotConfigured, empty file → real error, whitespace trimming
[...]

PR #22, opened 2026-04-01
config: add layered config commands (PRINFRA-141)

somanshreddy:
### What This Adds
This PR adds the first persistent config system for the CLI.

`heygen config set`, `heygen config get`, and `heygen config list` are backed by a local `config.toml` file in the CLI config directory. Config commands return structured JSON following the same stdout/stderr contract as the rest of the CLI. The config file is only created when the user explicitly runs `config set` — until then, all values come from defaults.

### Config Options

| Key | Type | Default | Env Override | Description |
|-----|------|---------|-------------|-------------|
| `output` | `json` \| `human` | `json` | `HEYGEN_OUTPUT` | Default output format |
| `analytics` | bool | `true` | `HEYGEN_NO_ANALYTICS` | PostHog analytics (opt-out) |

The `HEYGEN_NO_ANALYTICS` env var uses a negative naming convention (same as `NO_COLOR`, `HOMEBREW_NO_ANALYTICS`, `GH_NO_UPDATE_NOTIFIER`). Unset means default behavior (on); set to any non-empty value means disabled. This avoids ambiguity about whether an unset var means "on" or "off."

### Resolution Model
Config lookup uses a layered provider with clear precedence:

1. Environment variable
2. Config file (`~/.heygen/config.toml`)
3. Built-in default

### Type Handling
String settings (`output`) are stored as TOML strings. Boolean settings (`analytics`) are stored as TOML booleans. This prevents type drift between the CLI and on-disk config.

### Error Handling
A missing config file falls back to defaults silently. A corrupt or unreadable config file surfaces an error immediately — important for debugging persisted local settings.

### Example Outputs

**List all config (fresh install, no config file):**
```
$ heygen config list
[
  {"key": "analytics", "value": "true",  "source": "default"},
  {"key": "output",    "value": "json",  "source": "default"}
]
```

**Set a value:**
```
$ heygen config set output human
{"key": "output", "value": "human"}
```

**Get shows source (file vs env vs default):**
```
$ heygen config get output
{"key": "output", "value": "human", "source": "file"}

$ HEYGEN_OUTPUT=json heygen config get output
{"key": "output", "value": "json", "source": "env"}
```

**Validation errors:**
```
$ heygen config set bogus value
{"error": {"code": "usage_error", "message": "invalid config key \"bogus\""}}

$ heygen config set output xml
{"error": {"code": "usage_error", "message": "output must be one of: json, human"}}
```

**Linear issue:** PRINFRA-141

### Automated (25+ tests)
- **EnvProvider** (5): BaseURL default, Output default + custom, Analytics default + disabled, GetEnv mapping + bogus key
- **FileProvider** (11): Get exists/missing/no-file/corrupt, Set new-file/update-existing/overwrite/boolean-type/string-type, GetAll, GetAll corrupt
- **LayeredProvider** (5): all defaults, env overrides default, file overrides default, env overrides file, analytics default/from-file/env-disable
- **Config commands** (9): set success/invalid-key/invalid-value/api-base-rejected/skipAuth, get default/from-env/from-file/invalid-key, list all-defaults/mixed-sources
- All existing M0 + auth tests pass unchanged

### Manual smoke tests
- `heygen config list` → 2 keys with defaults
- `heygen config set output human` → TOML file written, `config get` shows `source: "file"`
- `HEYGEN_OUTPUT=json heygen config get output` → env overrides file, `source: "env"`
- `heygen config set analytics false` → stored as TOML boolean `false` (not string)
- `heygen config set bogus value` → exit 2
- `heygen config set api_base "https://..."` → exit 2 (not a user-facing key)

### DIFF
--- a/cmd/heygen/config_cmd.go
+++ b/cmd/heygen/config_cmd.go
@@ -57,7 +57,7 @@ func validateConfigValue(key, value string) error {
 		if value != "json" && value != "human" {
 			return clierrors.NewUsage("output must be one of: json, human")
 		}
-	case config.KeyAnalytics, config.KeyAutoUpdate:
+	case config.KeyAnalytics:
 		if value != "true" && value != "false" {
 			return clierrors.NewUsage(key + " must be one of: true, false")
 		}
--- a/cmd/heygen/config_cmd.go
+++ b/cmd/heygen/config_cmd.go
@@ -3,7 +3,6 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"net/url"
 	"slices"
 
 	"github.com/heygen-com/heygen-cli/internal/config"
@@ -61,11 +60,6 @@ func validateConfigValue(key, value string) error {
 	case config.KeyAnalytics, config.KeyAutoUpdate:
 		if value != "true" && value != "false" {
 			return clierrors.NewUsage(key + " must be one of: true, false")
-		}
-	case config.KeyAPIBase:
-		u, err := url.ParseRequestURI(value)
-		if err != nil || (u.Scheme != "http" && u.Scheme != "https") {
-			return clierrors.NewUsage("api_base must be a valid http or https URL")
 		}
 	}
 	return nil
--- a/cmd/heygen/config_cmd.go
+++ b/cmd/heygen/config_cmd.go
@@ -37,6 +37,14 @@ func configProviderWithSource(ctx *cmdContext) (config.ProviderWithSource, error
 	return provider, nil
 }
 
+func writableConfigProvider(ctx *cmdContext) (config.WritableProvider, error) {
+	provider, ok := ctx.configProvider.(config.WritableProvider)
+	if !ok {
+		return nil, clierrors.New("config provider does not support writes")
+	}
+	return provider, nil
+}
+
 func validateConfigKey(key string) error {
 	if !slices.Contains(config.ValidKeys, key) {
 		return clierrors.NewUsage(fmt.Sprintf("invalid config key %q", key))
@@ -87,11 +95,11 @@ func newConfigSetCmd(ctx *cmdContext) *cobra.Command {
 				return err
 			}
 
-			provider, ok := ctx.configProvider.(*config.LayeredProvider)
-			if !ok {
-				return clierrors.New("config provider does not support writes")
+			provider, err := writableConfigProvider(ctx)
+			if err != nil {
+				return err
 			}
-			if err := provider.File.Set(key, value); err != nil {
+			if err := provider.Set(key, value); err != nil {
 				return clierrors.New(err.Error())
 			}
 
--- a/cmd/heygen/config_cmd.go
+++ b/cmd/heygen/config_cmd.go
@@ -1,0 +1,171 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"slices"
+
+	"github.com/heygen-com/heygen-cli/internal/config"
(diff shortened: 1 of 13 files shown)


somanshreddy:
* **#22** <a href="https://app.graphite.com/github/pr/heygen-com/heygen-cli/22?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/heygen-com/heygen-cli/22?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#17** <a href="https://app.graphite.com/github/pr/heygen-com/heygen-cli/17?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

jrusso1020:
{"key": "api_base",    "value": "https://api.heygen.com", "source": "default"},  
  
would anyone besides us ever want to set this? I feel like we shouldn't expose this/allow users to proxy this  
  
{"key": "analytics", "value": "unset", "source": "default"},  
analytics should be opt out  
  
{"key": "output", "value": "json", "source": "default"}  
I'm not sure we want user to control this eitehr it should always be updates on no? user can --pin or something if they wish

jrusso1020:
why toml vs yaml or json?

somanshreddy:
@jrusso1020 

1. Hidden the API_BASE url from external users. It can be set by devs only.
2. So we want to turn it on by default but still let users opt out if they choose to, right?
3. I presume you are talking about `auto_update`? So should we remove the capability to toggle this via config?

somanshreddy:
*From Claude:*

@jrusso1020 TOML. The config file has 3 flat key-value pairs with booleans and strings — TOML's sweet spot.

YAML has implicit type coercion that's dangerous for config files with boolean keys: `no` silently becomes `false`, `on` becomes `true`, and the string `NO` (Norway country code) gets cast to a boolean. With keys like `analytics` and `auto_update`, that's a real footgun.

JSON doesn't support comments, so users can't annotate their config.

Stripe CLI uses TOML for the same reasons. `gh` uses YAML (but their config is mostly strings, fewer booleans). AWS uses INI.

jrusso1020:
1. sounds good yes devs only
2. Correct, opt out by default. So on to start but they can toggle off
3. yeah I don't think auto update should be configurable but I think you should be able to "pin" it per command invocation but not turn it off completely

jrusso1020 (approved):

somanshreddy:
Updated in `d01a042`:

- **auto_update removed** from config entirely. Version pinning will be a per-invocation flag (`--pin`) in future work.
- Config now has 2 keys: `output` and `analytics`
- TOML: sticking with it — avoids YAML's implicit type coercion (`no` → `false`, Norway problem), and supports comments unlike JSON. Same choice as Stripe CLI.

somanshreddy:
### Merge activity

* **Apr 2, 4:24 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/heygen-com/heygen-cli/22).
* **Apr 2, 4:28 AM UTC**: [Graphite](https://app.graphite.com/github/pr/heygen-com/heygen-cli/22) rebased this pull request as part of a merge.
* **Apr 2, 4:29 AM UTC**: @somanshreddy merged this pull request with [Graphite](https://app.graphite.com/github/pr/heygen-com/heygen-cli/22).

PR #49, opened 2026-04-06
scripts: rewrite installer for versioned releases

somanshreddy:
## Description

Replaces the dev-phase installer and release workflow with a versioned release cycle. Three changes:

**1. Dev release workflow rewrite (`.github/workflows/dev-release.yml`):**
The old workflow used a rolling `dev` tag with unversioned assets (`heygen_linux_amd64.tar.gz`) and `version="dev-<sha>"`. The new workflow auto-computes a semver-compliant tag from the latest stable release: bumps patch, appends `-dev.YYYYMMDD.SHORTHASH` (e.g., `v0.1.1-dev.20260406.abc1234`). Creates an immutable git tag and delegates to GoReleaser for versioned asset names, checksums, and prerelease publishing. Each dev build is a separate prerelease, not a rolling overwrite.

Version derivation: reads latest stable tag via `git tag --list`, bumps patch (`v0.1.0` → `v0.1.1`), appends dev metadata. If no stable tag exists, starts from `v0.0.1-dev.*`.

**2. Install script rewrite (`scripts/install.sh`):**
The old installer defaulted to the `dev` tag and required `GITHUB_TOKEN`. The new installer:
- Defaults to latest stable release (not dev)
- `--dev` flag installs latest dev prerelease
- `--version v0.1.0` pins to a specific version (`v` prefix required)
- Falls back to unauthenticated GitHub API when no token is available (ready for public repo)
- Uses versioned GoReleaser asset names (`heygen_v0.1.0_linux_amd64.tar.gz`)
- `--help` flag for discoverability
- `--dev` and `--version` are mutually exclusive

**3. GoReleaser asset naming (`.goreleaser.yaml`):**
Archive name template changed from `{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}` to `{{ .ProjectName }}_v{{ .Version }}_{{ .Os }}_{{ .Arch }}` — includes the `v` prefix in the filename for consistency with the tag format.

README, RELEASE.md, and SKILL.md updated to reflect the new install commands and versioning scheme.

Stacked on PR #23 (stable release workflow).

## Testing

Install script tested manually with `--help`, `--dev`, `--version`, and default (stable) modes. Dev release workflow validated by reading the computed version step output. GoReleaser config validated via `make goreleaser-check`.

PR #23, opened 2026-04-01
ci: add stable release workflow

somanshreddy:
## Description

Adds a manual GitHub Actions workflow for cutting stable releases. This is the first PR in the M6 release cycle stack (#23 → #49 → #50).

### Versioning Strategy

The CLI uses semver with a `v` prefix everywhere — git tags, `--version` output, `--version` flag input, JSON responses, install script.

| Build type | Tag format | Example | How triggered |
|---|---|---|---|
| **Stable** | `v{major}.{minor}.{patch}` | `v0.1.0` | Manual — GitHub Actions UI, operator inputs version |
| **Dev** | `v{base}-dev.{YYYYMMDD}.{shorthash}` | `v0.1.1-dev.20260406.abc1234` | Manual — GitHub Actions UI, version auto-computed |
| **Local** | — | `dev` | `go build` without ldflags |

**Semver ordering:** `v0.2.0` > `v0.1.1-dev.20260407.def` > `v0.1.1-dev.20260406.abc` > `v0.1.0`. Stable always beats prerelease. Dev builds sort chronologically by date.

**Dev version auto-derivation (PR #49):** The dev workflow reads the latest stable tag, bumps patch (`v0.1.0` → `v0.1.1`), and appends `-dev.YYYYMMDD.SHORTHASH`. No VERSION file, no manual bumps. Dev builds always sort above the latest stable.

**Bumping rules:** Pre-1.0 (`v0.x.y`): minor = features, patch = fixes, no stability guarantees. Post-1.0: full semver contract.

### This PR — Stable Release Workflow

**Trigger:** GitHub Actions UI → "Stable Release" → input version tag (e.g., `v0.1.0`) → "Run workflow"

**What it does:**
1. Validates version is semver format (`v0.1.0` — rejects `1.0` or `foo`)
2. Validates the tag doesn't already exist (prevents accidental overwrite)
3. Runs tests
4. Creates git tag on current `main` HEAD
5. GoReleaser builds platform binaries + `checksums.txt`, publishes as a stable GitHub release

**Downstream PRs:**
- PR #49 — rewrites dev release workflow + installer for versioned assets
- PR #50 — adds `heygen update` / `heygen update check` commands (channel-aware: stable users get stable updates, dev users get dev updates)

## Testing

Workflow syntax validated. GoReleaser config tested by `goreleaser-check` CI job. Version validation tested manually with valid/invalid inputs.

PR #50, opened 2026-04-06
cmd: add heygen update command

somanshreddy:
## Description

Adds `heygen update` and `heygen update check` — explicit self-update commands for the CLI. No auto-update, no startup notifications, no network calls in `--help` or `--version`.

**`heygen update check`** — queries GitHub Releases for the latest stable version, returns JSON:
```json
{"current": "v0.1.0", "latest": "v0.2.0", "update_available": true}
```

**`heygen update`** — downloads the latest stable release, verifies SHA256 checksum, replaces the binary. Returns JSON:
```json
{"previous": "v0.1.0", "current": "v0.2.0", "message": "Updated heygen from v0.1.0 to v0.2.0"}
```

**`heygen update --version v0.1.0`** — pins to a specific version. The `v` prefix is required; bare `0.1.0` returns exit 2 (usage error).

**Design decisions:**
- `skipAuth: true` — no HeyGen API key needed. GitHub Releases are the source.
- GitHub token sourced from `GITHUB_TOKEN` → `GH_TOKEN` → `gh auth token` (same chain as the installer). Empty token works once the repo is public.
- Package manager detection: if binary is under `/homebrew/`, `/Cellar/`, `/linuxbrew/`, or `node_modules/`, returns a `CLIError` telling the user to use their package manager instead.
- Local `dev` builds (no ldflags) get a clear error: "reinstall from a release build to use heygen update."
- `releaseUpdater` interface abstracts `go-selfupdate` for testability. All tests use a mock — no real GitHub API calls.
- Static hint in root `--help`: `Use "heygen update check" to see if a newer version is available.` No dynamic notifications.
- `version` field added to `cmdContext` to pass the build-time version string to the update command.

**Dependencies added:** `creativeprojects/go-selfupdate` (GitHub Releases source, checksum validation, atomic binary replacement, rollback on failure) and `Masterminds/semver` (version parsing and validation).

Stacked on PR #49 (installer rewrite).

## Testing

7 tests with mock `releaseUpdater`:
- `TestUpdateCheck_NewVersionAvailable` — newer version returns `update_available: true`
- `TestUpdateCheck_AlreadyCurrent` — same version returns `update_available: false`
- `TestUpdateCheck_LocalDevVersion` — `dev` build returns exit 1 with "not release-tagged" error
- `TestUpdate_SkipsAuth` — no API key needed, exit 0
- `TestUpdate_PackageManagerWarning` — Homebrew path returns `wrong_install_method` error with brew hint
- `TestUpdate_JSONOutput` — success returns valid JSON with previous/current/message
- `TestUpdate_RejectsBareVersion` — `--version 0.1.0` (no `v`) returns exit 2

Root help test updated to verify the static update hint line.


somanshreddy:
* **#50** <a href="https://app.graphite.com/github/pr/heygen-com/heygen-cli/50?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/heygen-com/heygen-cli/50?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#49** <a href="https://app.graphite.com/github/pr/heygen-com/heygen-cli/49?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23** <a href="https://app.graphite.com/github/pr/heygen-com/heygen-cli/23?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

README.md
```diff
@@ -40,6 +40,13 @@ gh api repos/heygen-com/heygen-cli/contents/scripts/install.sh \
   --jq '.content' | base64 -d | bash -s -- --version v0.1.0
 ```
 
+Check for and apply updates after install:
+
+```bash
+heygen update check
```
jrusso1020: does this need to be a separate command or a flag? just curious if we have examples here of  
  
I coudl see `heygen version` being waht does the check vs heygen update does check + downloads

cmd/heygen/root.go
```diff
@@ -83,6 +86,7 @@ func newRootCmdWithSpecs(version string, formatter output.Formatter, groups map[
 
 	root.AddCommand(newAuthCmd(ctx))
 	root.AddCommand(newConfigCmd(ctx))
+	root.AddCommand(newUpdateCmd(ctx))
```
jrusso1020, line 89: do we not have a version command?

README.md
somanshreddy: So, Cobra autogenerates the `--version`/`-v` flag on `heygen`.

We can add an explicit version subcommand as well. Though I think `-v` is a good standard.

Also, a quick comparison table on having version separate from an explicit update check:

somanshreddy: Initially, I wanted `--version` to also check for an update. Both codex and claude pushed back saying that this is usually a non-networked call. 

But I still feel that it might be useful. What do you think? Running `--version` also checks if there is a newer version available. However, we would need to see how agents react to this behaviour (version flag for checking for a new version and then update to actually update the package)

somanshreddy: @jrusso1020 The passive notice pattern above is below: 

> When you run any gh or stripe command, they occasionally print a message to stderr like:
>
> A new release of gh is available: 2.44.0 → 2.45.0
> https://github.com/cli/cli/releases/tag/v2.45.0
>
> It's "passive" because:
>
> - You didn't ask for it — it just appears alongside your normal command output
> - It's on stderr so it doesn't break JSON piping
> - It has a 24-hour cooldown so it doesn't show every time
> - It's informational only — you still have to run brew upgrade gh yourself

cmd/heygen/update.go
```diff
@@ -0,0 +1,314 @@
...
+
+func (u *selfUpdater) UpdateTo(ctx context.Context, rel updateRelease, cmdPath string) error {
+	return u.updater.UpdateTo(ctx, rel.raw, cmdPath)
+}
+
+var (
+	newReleaseUpdater = func(prerelease bool) (releaseUpdater, error) {
+		source, err := selfupdate.NewGitHubSource(selfupdate.GitHubConfig{
+			APIToken: githubReleaseToken(),
+		})
```
jrusso1020, line 68: if you don't have github account you can't download from here? is that a potential concern

somanshreddy: Only while the repo is private. This flow won't be used once the repo is public.

This code should already work for a public repo. An empty string will be returned if user doesn't have a GitHub account, and then it would proceed with an unauthenticated request to GitHub. It wouldn't throw an error when the repo is public.

README.md
jrusso1020: Yeah makes sense to me I think --version could be teh default cobra and we could add version command to if we wnat to have custom logic?   
  
I just find the update check command confusing/unnecessary can eitehr include that info with version or just have an update command

somanshreddy: Okay. We can simplify to the below for now:

> `heygen --version` — local version string (Cobra built-in, what everyone expects)
> `heygen update` — checks for newer version + installs if available. If already current, tells you.

We can add a passive note as a follow up feature.

cmd/heygen/root.go
somanshreddy: Keeping `heygen --version` (Cobra built-in) as the version display. No separate heygen version subcommand, the same as aws and rustup. 

If we need custom version logic later, we can add it then.

jrusso1020 (approved):

somanshreddy:
### Merge activity

* **Apr 7, 1:01 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/heygen-com/heygen-cli/50).
* **Apr 7, 1:01 AM UTC**: @somanshreddy merged this pull request with [Graphite](https://app.graphite.com/github/pr/heygen-com/heygen-cli/50).

PR #70, opened 2026-04-08
ci: verify generated commands against checked-in spec

somanshreddy:
## Description

Adds the CLI-side infrastructure for automated spec sync from experiment-framework.

**Three components:**

1. `codegen/external-api.json`: checked-in copy of the OpenAPI spec (matches current `gen/`). This is the source of truth for CLI codegen going forward.

2. `auto-generate-codegen.yml`: when the EF sync workflow pushes an updated spec to a `sync/cli-codegen-*` branch, this workflow runs `make generate` and pushes the regenerated `gen/` files. Uses `GIT_PAT` so the push triggers downstream CI.

3. `codegen-check` CI job (PRINFRA-147): on every PR, runs `make generate` against the checked-in spec and fails if `gen/` doesn't match. Catches stale codegen, hand-edits to `gen/`, and sync PRs where auto-generation failed.

**The full chain:**
1. EF spec changes on master
2. EF workflow copies spec to heygen-cli sync branch, opens PR with `product-infra` and `somanshreddy` as reviewers, assigned to the EF PR author (heygen-com/experiment-framework#34071)
3. `auto-generate-codegen.yml` runs codegen, pushes `gen/` to the PR
4. `codegen-check` verifies sync, CI passes
5. Team reviews and merges

**Prerequisites:**
- `GIT_PAT` must be added as a repo secret on `heygen-com/heygen-cli` (same token value as EF). Needed by `auto-generate-codegen.yml` to push generated code and trigger CI. Without it, the auto-generate workflow cannot push back to the sync branch.

**Companion PR:** heygen-com/experiment-framework#34071 (EF-side workflow)

## Testing

- `codegen-check` passes on this PR (spec matches current `gen/`)
- `make generate SPEC=codegen/external-api.json` produces no diff
- Auto-generate workflow validated via YAML syntax
- End-to-end testing requires both PRs merged and an EF spec change


somanshreddy:
Superseded by the revised sync design. CLI-side spec storage, auto-generation, and codegen-check are no longer needed; the EF workflow now runs codegen directly and opens sync PRs containing only regenerated gen/ files.

PR #125, opened 2026-04-27
errors: add retryable field to CLIError for machine consumers

somanshreddy:
## Description

PostHog analytics (Apr 21-27) revealed a polling storm pattern: a single install called
`lipsync get` 5,429 times with a bad ID. The CLI's built-in `--wait` polling has backoff,
but external scripts and LLM agents retrying not-found errors had no signal that these
errors are permanent.

This PR adds a `retryable` field to the error JSON envelope. It is best-effort tri-state
guidance for machine consumers:

- **`true`**: transient; same request may succeed later (429, 5xx, network errors)
- **`false`**: permanent; retrying won't help without changing input (known not-found codes, auth errors, insufficient credit)
- **omitted**: unknown or context-dependent (generic 404 without recognized API code, validation errors, `--wait` timeouts, request deadline timeouts)

Classification is code-based first, HTTP status as fallback. A generic 404 without a
recognized API error code returns `retryable: nil` (unknown), not `false`.

### Hint and retryable consistency

Both hints and retryable use the raw API error code (`apiErr.Code`), not the display code
defaulted from HTTP status. This ensures the three user-facing surfaces (code, hint, retryable)
tell one story:

- Explicit `not_found` from API: permanence hint + `retryable: false`
- Generic 404 (empty API code): no hint + `retryable: nil`
- `asset_not_available`: "may still be processing" + `retryable: nil`

### Timeout semantics

`code: "timeout"` appears in three contexts with different `retryable` values:
- API timeout error code: `retryable: true` (server-side operation still processing)
- `--wait` CLI timeout: `retryable: nil` (create succeeded but poll timed out; whether to re-poll or re-create depends on caller context)
- Request deadline exceeded: `retryable: nil` (unknown whether server received request)

SKILL.md documents that agents should branch on `retryable`, not `code == "timeout"`.

### Discoverability

The error envelope shape and retryable semantics are now documented in:
- `heygen --help` (Error Envelope section): agents discover this on first use
- `SKILL.md`: the full agent-facing contract
- `README.md`: user-facing docs

```json
{"error": {"code": "video_not_found", "message": "...", "hint": "...", "retryable": false}}
```

`retryable` is omitted entirely when unknown.

**Stacked on #124**

## Testing

- Retryable classification: permanent codes -> false, 500/429 -> true, unknown 404 -> nil, generic empty 404 -> nil + no permanence hint, auth -> false
- Envelope JSON shape: includes `retryable` when set, omits when nil
- `NewAuth` constructor sets `retryable: false`
- Network error sets `retryable: true`
- `make test` passes (Linux, macOS, Windows)

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Every command supports `--help`.
 | Aspect | Behavior |
 |--------|----------|
 | **stdout** | Always JSON. Even `video download` — binary writes to disk; stdout emits `{"asset", "message", "path"}` so you can chain on `.path`. |
-| **stderr** | Structured envelope on error: `{"error": {"code", "message", "hint", "retryable"}}`. Stable `code` values for programmatic branching. `retryable` is best-effort: `true` (transient), `false` (permanent), or omitted (unknown). |
+| **stderr** | Structured envelope on error: `{"error": {"code", "message", "hint", "retryable"}}`. Stable `code` values for programmatic branching. When present, `retryable` is definitive: `true` (transient), `false` (permanent). Omitted when unrecognized. |
 | **Exit codes** | `0` ok · `1` API or network · `2` usage · `3` auth · `4` timeout under `--wait` (stdout contains partial resource for resume) |
 | **Request bodies** | Flags for simple inputs; `-d` for nested JSON (inline, file path, or `-` for stdin). Flags override matching fields. |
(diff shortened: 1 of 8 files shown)


somanshreddy:
Dropping the retryable field. The hint text already communicates permanence/transience to agents in natural language, which is simpler and doesn't require a separate machine-readable tri-state contract. The valuable parts (permanence hints, raw API code for hint lookup) moved to PR #124.

PR #151, opened 2026-05-22
codegen: honor x-cli-default for per-surface flag defaults

ivanbezborodov-heygen:
## Description

context [thread](https://heygen.slack.com/archives/C08CG5T67JL/p1779217121124659)
The HTTP API and the CLI sometimes want different defaults for the same field. ``aspect_ratio`` is the immediate motivator: the API defaults to ``16:9`` for backwards compatibility, but agent-driven CLI flows are better off defaulting to ``auto`` so the canvas tracks the source orientation by default.

EF authors signal this from the Pydantic field via ``json_schema_extra={"x-cli-default": "auto"}``, which lands verbatim on the schema property in ``openapi/external-api.json``. This change teaches the CLI codegen to prefer ``x-cli-default`` over ``default`` when present.

Mechanics:

- New ``schemaCliDefault`` helper mirrors the style of ``isCliVisible`` / ``isCliAction``. It returns ``(value, ok)`` so callers can distinguish "no default at all" from "explicit null default".
- Both default-reader sites in ``GroupEndpoints`` (query params, JSON body fields) route through the helper. The else-arm preserves the prior ``s.Default``/``prop.Default`` behavior, so this is a no-op on any spec that doesn't carry ``x-cli-default``.

``gen/`` is intentionally untouched in this PR. The auto-sync workflow will regenerate it after the EF PR carrying the ``x-cli-default`` override merges, and the next sync output will flip ``--aspect-ratio``'s default to ``auto`` automatically.
## Testing

- Tests:
  - ``TestSchemaCliDefault`` table-drives the precedence rules (extension wins over default, missing both returns ok=false, x-mcp-default is intentionally ignored).
  - ``TestGroupEndpoints_BodyFlagsRespectXCliDefault`` runs end-to-end on a test spec with both an override field (``aspect_ratio``) and a fallback control (``fps``).
- Test spec gains those two fields on the ``POST /v3/videos`` body.


somanshreddy (changes requested):
Blocking on the CLI default behavior: this currently only changes the generated Cobra flag default/help, not the request payload sent when the user omits the flag.

`schemaCliDefault` writes the extension value into `FlagSpec.Default` in `codegen/grouper.go`, and `registerFlag` uses that as the Cobra default. But `BuildInvocation` explicitly skips every flag where `cmd.Flags().Changed(flag.Name)` is false, so an omitted `--aspect-ratio` is not added to the body at all. The API then receives no `aspect_ratio` and applies its HTTP default (`16:9`), not the intended CLI default (`auto`).

So after EF regenerates `--aspect-ratio` with `Default: "auto"`, `heygen video create ...` without `--aspect-ratio` will still behave like API default `16:9`; only the help/default display changes.

I think the fix needs to distinguish API defaults from CLI override defaults, then materialize only CLI override defaults into the invocation even when the flag was not explicitly changed. Ordinary OpenAPI `default` values should probably keep the existing omit-unless-changed behavior so the CLI does not start sending every server default back to the API.

ivanbezborodov-heygen:
Addressed in 829d39a.

Root-causing the gap: `schemaCliDefault` was wiring the value into `FlagSpec.Default`, which `registerFlag` correctly translates into the Cobra flag default — but `BuildInvocation` then skips any flag where `cmd.Flags().Changed(...)` is `false`. So an omitted `--aspect-ratio` was never written into the request body, the server applied its own `16:9` default, and the CLI's `auto` was a help-text-only fiction. Your read was exactly right.

Fix:

- Added `FlagSpec.ForceSend bool`. It is set **only** when the default came from `x-cli-default` (the extension); ordinary OpenAPI `default` values leave it `false`, so they keep the existing omit-unless-changed behavior and the CLI does not start echoing every server default back to the server.
- `schemaCliDefault` now returns `(value, ok, fromExtension)`; both reader sites in `GroupEndpoints` propagate `fromExtension` into `flag.ForceSend`.
- `command.go.tmpl` emits `ForceSend: true,` when set.
- `BuildInvocation` short-circuit becomes `!Changed(...) && !flag.ForceSend`. Once the gate passes, the value is read from Cobra's store exactly like the user-supplied case — so a user-supplied value still wins naturally; ForceSend just keeps the gate open when the user is silent.

Tests covering the matrix:

- `internal/command`:
  - `ForceSendWritesBodyDefaultWhenFlagOmitted` — the core fix: silent user, body carries `aspect_ratio=auto`.
  - `ForceSendUserValueWins` — explicit `--aspect-ratio 9:16` still beats the ForceSend default.
  - `ForceSendForQueryParam` — symmetry with body sources.
  - `NonForceSendDefaultStaysOmitted` — regression guard against the "echo every server default" anti-pattern.
  - `ForceSendValidatesAgainstEnum` — defends against a future codegen bug where the `x-cli-default` value falls outside the enum.
- `codegen`:
  - `TestSchemaCliDefault` gains a `wantFromExtension` column.
  - `TestGroupEndpoints_BodyFlagsRespectXCliDefault` asserts `ForceSend=true` on `aspect_ratio` (override) and `ForceSend=false` on `fps` (default-only control).

Local `go test ./...` and `go vet ./...` green.

somanshreddy:
I found one additional bug in the `ForceSend` implementation.

`ForceSend` currently keeps the flag-processing gate open whenever the user omits a flag with an `x-cli-default`:

```go
if !cmd.Flags().Changed(flag.Name) && !flag.ForceSend {
    continue
}
```

That fixes the omitted-flag path, but it also means an omitted generated flag can overwrite an explicit request body supplied via `--data`.

Example shape:

```bash
heygen video create --data {"aspect_ratio":"16:9", ...}
```

With the current code, `inv.Body` starts from `--data`, then the omitted `--aspect-ratio` flag has `ForceSend=true`, so `BuildInvocation` writes Cobra default `auto` into `inv.Body["aspect_ratio"]`. The request sent to the API becomes `auto`, even though the user explicitly provided `16:9` in the JSON body.

Suggested fix: keep the `ForceSend` behavior for the no-body / missing-field case, but make it a non-destructive default fill when the flag itself was not changed:

```go
changed := cmd.Flags().Changed(flag.Name)
if !changed && !flag.ForceSend {
    continue
}

if flag.Source == "body" && !changed && flag.ForceSend && inv.Body != nil {
    if _, exists := inv.Body[flag.JSONName]; exists {
        continue
    }
}
```

This preserves the intended CLI default behavior:

- no `--aspect-ratio`, no `--data.aspect_ratio` -> send `aspect_ratio=auto`
- explicit `--aspect-ratio 9:16` -> send `9:16`
- explicit `--data {"aspect_ratio":"16:9"}` -> keep `16:9`

I would also consider renaming `ForceSend` to something closer to the semantics, like `SendDefaultWhenOmitted`, but the overwrite fix is the important part.

somanshreddy (approved):
👍 with one comment regarding an edge case

ivanbezborodov-heygen:
Addressed in d1871cf.

Excellent catch — you were exactly right. The previous commit kept the gate open for any unchanged flag with `ForceSend=true`, which silently rewrote explicit user values supplied via `-d/--data` JSON. That's worse than the original bug because the user said `16:9` and the API received `auto`.

Fix:

- Added a field-scoped existence check in `BuildInvocation`: when the user didn't pass the flag, the source is `body`, the default-fill gate is the only reason we're processing it, AND the body already carries a value for the same JSON field (typically from `--data`), skip the write. Query params and file uploads are untouched because `--data` is body-only.
- Final precedence is now: `explicit --aspect-ratio` > `--data {"aspect_ratio": ...}` > CLI default.
- Renamed `ForceSend` → `SendDefaultWhenOmitted` as suggested. The new name describes the trigger condition rather than implying unconditional override, which makes the non-destructive semantics easier to read at the call site.

Three new tests pin the contract:

- `SendDefaultWhenOmittedDoesNotOverrideDataJSON` — the core regression: `--data` carries `aspect_ratio="16:9"`, `--aspect-ratio` omitted; body ends up `16:9`, not `auto`. Title from `--data` is still preserved (sanity check that the guard is field-scoped, not blanket).
- `SendDefaultWhenOmittedFillsMissingDataField` — `--data` is present but doesn't include `aspect_ratio`; the default fill still applies for that missing field.
- `SendDefaultWhenOmittedExplicitFlagOverridesDataJSON` — three-way precedence locked.

Existing five tests renamed for consistency; nothing else changed.

(Small side note: 6fe3250 drops some `.idea/` files my `git add -A` accidentally swept in on the previous push, and gitignores `.idea/` / `.vscode/` so it can't recur.)

`go test ./...` + `go vet ./...` green.

PR #154, opened 2026-05-26
auth: read JSON credential format (read-side backwards compat)

jrusso1020:
## Description

Teach heygen-cli's `FileCredentialResolver` to read the new JSON
credentials format that hyperframes-CLI writes, while staying fully
backwards-compatible with the legacy single-line plaintext file.

The shared file at `~/.heygen/credentials` may now contain either:

1. **JSON layout** (new):
   ```json
   {
     "api_key": "hg_...",
     "oauth": {
       "access_token": "...",
       "refresh_token": "...",
       "expires_at": "2026-06-25T12:00:00Z"
     }
   }
   ```
   Unexpired `oauth.access_token` wins over `api_key`; expired
   OAuth falls through to `api_key` (heygen-cli doesn't refresh
   today — that lives in hyperframes-CLI).
2. **Legacy plaintext** (single-line API key): kept readable so
   existing users don't lose their session.

The public `FileCredentialResolver.Resolve() (string, error)`
signature is unchanged — callers continue to treat the returned
string as opaque. When heygen-cli gains a Bearer-aware HTTP client,
the resolver can be expanded to return a typed credential.

Implementation notes:

- JSON detection is `trimmed[0] == '{'` — anything else routes to
  the legacy path.
- A 60s expiry skew matches the hyperframes-CLI resolver
  (`packages/cli/src/auth/resolver.ts`) so both CLIs agree on
  staleness.
- Unparseable `expires_at` is treated as fresh — the server is the
  source of truth for token validity.
- Unknown JSON fields are silently dropped (forward compat).
- Multi-line non-JSON garbage is rejected explicitly (previously
  would have been returned as a "credential").

## Testing

- 15 unit tests in `internal/auth/file_resolver_test.go` (4 legacy
  cases preserved, 11 new JSON cases): api_key only, oauth only
  fresh, both-fresh-oauth-wins, expired-oauth-fallback, expired-
  no-api-key error, no-expires_at lenient, invalid JSON, empty
  object, multi-line garbage rejected, unparseable expires_at,
  unknown-field drop.
- `go test ./internal/auth/...` clean.
- `go vet ./internal/auth/...` clean.
- Manual smoke (recommended pre-merge): `hyperframes auth login --api-key`
  on the prior hyperframes PR, then `heygen video list --limit 1`
  in this build — both CLIs share the same file with no extra
  setup.

Pre-existing test failure on `main` (`TestVideoList_AuthMissing`)
is unrelated to this PR — verified by running on a clean
`origin/main` checkout.

### DIFF
--- a/internal/auth/file_resolver.go
+++ b/internal/auth/file_resolver.go
@@ -1,20 +1,56 @@
 package auth
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/heygen-com/heygen-cli/internal/paths"
 )
 
+// jsonCredentials is the on-disk format hyperframes-CLI (and future
+// heygen-cli releases) write to ~/.heygen/credentials. Mirror the
+// shape used by packages/cli/src/auth/store.ts in hyperframes-oss.
+type jsonCredentials struct {
+	APIKey string           `json:"api_key,omitempty"`
+	OAuth  *jsonOAuthTokens `json:"oauth,omitempty"`
+}
+
+type jsonOAuthTokens struct {
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token,omitempty"`
+	// ExpiresAt is ISO-8601 UTC. The hyperframes resolver allows a 60s
+	// skew; we use the same here so behavior matches across CLIs.
+	ExpiresAt string `json:"expires_at,omitempty"`
+	Scope     string `json:"scope,omitempty"`
+	TokenType string `json:"token_type,omitempty"`
+}
+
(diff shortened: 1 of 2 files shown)


somanshreddy:
I re-reviewed this parent PR as well. I’m not approving this one standalone because its current head still prefers `oauth.access_token` over `api_key`, while the client only sends the resolved string as `x-api-key`. That specific issue is fixed in the child PR (#155), where `api_key` wins and OAuth-only files are rejected, so the final stack looks good once #155 is included.

If this parent is intended to merge independently, it should get the same credential-selection fix from #155 before landing.

somanshreddy (approved):
Approving with the understanding that this lands together with #155. The parent PR's temporary OAuth-selection behavior is corrected by the child PR, and the final stack is in good shape.

PR #168, opened 2026-06-11
output: render nested objects in --human as an indented, humanized layout (PRINFRA-153)

somanshreddy:
## Scope

**Surfaces:** CLI | **Module:** Output/Formatter

## Description

`--human` renders single-object API responses as key-value output. A nested object value used to dump as a raw JSON blob (e.g. `heygen auth status --human` showed `wallet` as `{"auto_reload":{"enabled":false},...}`). PRINFRA-153 asks for a readable layout.

This rewrites the single-object path to an **indented, humanized layout similar to `kubectl describe`**. (An interim revision of this PR flattened nesting into dotted keys like `wallet.auto_reload.enabled`; per review we moved to the indented style, which is consistent with the table path — already humanized — and reads better than humanized dotted paths.)

```
Status:  completed
Wallet:
  Auto Reload:
    Enabled:  false
  Currency:           usd
  Remaining Balance:  476.78
Messages:
  - Content:  hello
    Role:     user
```

Approach:
- **Humanized labels per segment** via the existing `humanizeKey` (`auto_reload` -> "Auto Reload"), matching the table headers so list and detail views read consistently.
- **Nested objects** render as a `Label:` header, then recurse at +2 spaces of indent.
- **Per-level local alignment:** scalar siblings align to the longest label *within their own block*, not a single global column, so one deep key no longer pushes every value far right.
- **Arrays of objects** render as YAML-style `- ` sequence items; **homogeneous scalar arrays** join inline (`a, b, c`); **mixed-type or null-containing arrays** fall back to compact JSON (preserving type fidelity); **empty objects/arrays and `null`** render as `(none)`.
- Leaf formatting (Unix-timestamp dates, durations, multiline strings) is preserved; keys are sorted at every level (deterministic output).

JSON output (the default formatter) and the table (array-of-objects) path are untouched. The README also documents `--human` as a may-change presentation layer — scripts and agents should consume the stable default JSON.

## Testing

`make build` and `make test` pass. The key-value unit tests were rewritten to assert the exact indented output (nested object, deep nesting, arrays of objects with `- ` items, per-level alignment, scalar-array inline, mixed/null -> compact JSON, empty collections and `null` -> `(none)`, multiline continuation, depth-guard compact-JSON fallback). Two `--human` assertions outside the package (`generated_root_test.go`, `builder_test.go`) were updated to the humanized labels. The `lint` CI check is green (CI runs a go1.25-compatible golangci-lint). `make lint` fails only on the local devbox, whose installed golangci-lint is built for go1.23 and refuses the go1.25 module — an environment issue, not a code one; `gofmt -l` is clean and `go vet ./internal/output/ ./cmd/heygen/` passes. Reviewed with Codex (4 rounds) plus a PR review that caught a sequence-item nested-first-field indent bug and a null-at-depth-cap inconsistency, both now fixed with tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ heygen video download <video-id>
 
 Add `--human` to any command for a readable layout. Set `HEYGEN_OUTPUT=human` to make it the default.
 
-In `--human` mode, list responses render as a table and single-object responses render as an indented, humanized layout (similar to `kubectl describe`). Keys are humanized per segment (`auto_reload` → `Auto Reload`); nested objects indent under a `Label:` header, scalar siblings align locally within each block, arrays of scalars join inline (`a, b, c`), and arrays of objects render as YAML-style `- ` sequence items. Empty objects, empty arrays, and nulls show as `(none)`. Long scalar values (e.g. pre-signed download URLs) are printed in full and may wrap in a narrow terminal; pipe the default JSON to `jq` if you need to extract one cleanly. For example:
+In `--human` mode, list responses render as a table and single-object responses render as an indented, humanized layout (similar to `kubectl describe`). Keys are humanized per segment (`auto_reload` → `Auto Reload`), with common acronyms uppercased (`video_url` → `Video URL`); nested objects indent under a `Label:` header, scalar siblings align locally within each block, arrays of scalars join inline (`a, b, c`), and arrays of objects render as YAML-style `- ` sequence items. Empty objects, empty arrays, and nulls show as `(none)`. Long scalar values (e.g. pre-signed download URLs) are printed in full and may wrap in a narrow terminal; pipe the default JSON to `jq` if you need to extract one cleanly. For example:
 
 ```
 Status:  completed
--- a/README.md
+++ b/README.md
@@ -98,7 +98,19 @@ heygen video download <video-id>
 
 Add `--human` to any command for a readable layout. Set `HEYGEN_OUTPUT=human` to make it the default.
 
-In `--human` mode, list responses render as a table and single-object responses render as aligned key-value rows. Nested objects are flattened into dotted keys (e.g. `wallet.auto_reload.enabled: false`); arrays of scalars are joined inline (`a, b, c`) and arrays of objects are indexed (`messages.0.role`). JSON output (the default) is never altered.
+In `--human` mode, list responses render as a table and single-object responses render as an indented, humanized layout (similar to `kubectl describe`). Keys are humanized per segment (`auto_reload` → `Auto Reload`); nested objects indent under a `Label:` header, scalar siblings align locally within each block, arrays of scalars join inline (`a, b, c`), and arrays of objects render as YAML-style `- ` sequence items. Empty objects, empty arrays, and nulls show as `(none)`. For example:
+
+```
+Status:  completed
+Wallet:
+  Auto Reload:
+    Enabled:  false
+  Currency:           usd
+  Remaining Balance:  476.78
+Tags:  alpha, beta
+```
+
+`--human` is a readable layout for terminals and may change between releases; scripts and agents should consume the default JSON output, which is stable. JSON output (the default) is never altered.
 
 ## Commands
 
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ heygen video download <video-id>
 
 Add `--human` to any command for a readable layout. Set `HEYGEN_OUTPUT=human` to make it the default.
 
-In `--human` mode, list responses render as a table and single-object responses render as aligned key-value rows. Nested objects are flattened into dotted keys (e.g. `wallet.auto_reload.enabled  false`); arrays of scalars are joined inline (`a, b, c`) and arrays of objects are indexed (`messages.0.role`). JSON output (the default) is never altered.
+In `--human` mode, list responses render as a table and single-object responses render as aligned key-value rows. Nested objects are flattened into dotted keys (e.g. `wallet.auto_reload.enabled: false`); arrays of scalars are joined inline (`a, b, c`) and arrays of objects are indexed (`messages.0.role`). JSON output (the default) is never altered.
 
 ## Commands
 
--- a/README.md
+++ b/README.md
@@ -98,6 +98,8 @@ heygen video download <video-id>
 
 Add `--human` to any command for a readable layout. Set `HEYGEN_OUTPUT=human` to make it the default.
 
+In `--human` mode, list responses render as a table and single-object responses render as aligned key-value rows. Nested objects are flattened into dotted keys (e.g. `wallet.auto_reload.enabled  false`); arrays of scalars are joined inline (`a, b, c`) and arrays of objects are indexed (`messages.0.role`). JSON output (the default) is never altered.
+
 ## Commands
 
 Mirrors the [HeyGen v3 API](https://developers.heygen.com). Pattern: `heygen <noun> <verb>`.
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ heygen video download <video-id>
 
 Add `--human` to any command for a readable layout. Set `HEYGEN_OUTPUT=human` to make it the default.
 
-In `--human` mode, list responses render as a table and single-object responses render as an indented, humanized layout (similar to `kubectl describe`). Keys are humanized per segment (`auto_reload` → `Auto Reload`); nested objects indent under a `Label:` header, scalar siblings align locally within each block, arrays of scalars join inline (`a, b, c`), and arrays of objects render as YAML-style `- ` sequence items. Empty objects, empty arrays, and nulls show as `(none)`. For example:
+In `--human` mode, list responses render as a table and single-object responses render as an indented, humanized layout (similar to `kubectl describe`). Keys are humanized per segment (`auto_reload` → `Auto Reload`); nested objects indent under a `Label:` header, scalar siblings align locally within each block, arrays of scalars join inline (`a, b, c`), and arrays of objects render as YAML-style `- ` sequence items. Empty objects, empty arrays, and nulls show as `(none)`. Long scalar values (e.g. pre-signed download URLs) are printed in full and may wrap in a narrow terminal; pipe the default JSON to `jq` if you need to extract one cleanly. For example:
 
 ```
 Status:  completed
--- a/cmd/heygen/builder_test.go
+++ b/cmd/heygen/builder_test.go
@@ -605,7 +605,7 @@ func TestGenBuilder_VideoCreate_Wait_Human_NonTTYFallback(t *testing.T) {
 	if !strings.Contains(res.Stderr, "Polling: status=processing") {
 		t.Fatalf("stderr = %s, want plain-text non-TTY progress", res.Stderr)
 	}
-	if !strings.Contains(res.Stdout, "status:") {
+	if !strings.Contains(res.Stdout, "Status:") {
 		t.Fatalf("stdout = %s, want human-formatted output", res.Stdout)
 	}
 }
--- a/cmd/heygen/builder_test.go
+++ b/cmd/heygen/builder_test.go
@@ -605,7 +605,7 @@ func TestGenBuilder_VideoCreate_Wait_Human_NonTTYFallback(t *testing.T) {
 	if !strings.Contains(res.Stderr, "Polling: status=processing") {
 		t.Fatalf("stderr = %s, want plain-text non-TTY progress", res.Stderr)
 	}
-	if !strings.Contains(res.Stdout, "Status:") {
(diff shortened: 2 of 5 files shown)


james-russo-rames-d-jusso (commented):
*Approve-with-concerns* — the rewrite is clean and the scope is exactly what's advertised; I found one real rendering bug in array-of-objects whose first sorted key is itself a nested object/array, plus a couple of small consistency nits.

*What this PR ships:*
Rewrites `HumanFormatter.renderKeyValue` (the single-object `--human` path) from raw-JSON-blob-for-nested into a `kubectl describe`-style indented humanized layout. Three previous iterations in the commit log (Codex caught a null inconsistency). Per-level local alignment, `- ` YAML markers for arrays of objects, homogeneous-scalar inline join, mixed-scalar → compact JSON, empty/null → `(none)`. README updated to flag `--human` as a may-change presentation layer. JSON output, table path, primitive-list path, and `formatCell`/`formatValue` leaf formatting are all untouched — I diffed the file and confirmed.

*Backwards-compat assessment:*
- `--human` is opt-in and `HEYGEN_OUTPUT=human` requires an explicit env set; default formatter is still JSON. Script consumers are explicitly out of scope per the new README paragraph ("scripts and agents should consume the default JSON output, which is stable").
- The table (array-of-objects) path is untouched, so list views are unaffected.
- Anyone who *was* parsing `--human` single-object output (raw-JSON-blob for nested values, `[1]/[2]` markers for arrays of objects) will see different output. That's the point of the PR and the README disclaimer covers it.
- No ANSI / color behavior changes (the new key-value path emits pure ASCII, no `lipgloss` styling on values).

*Concerns:*

- *`internal/output/human_formatter.go:266-278` — array-of-objects with non-scalar first sorted key renders ambiguously.* When a sequence-item element's first sorted key is itself a non-empty nested object or array-of-objects, the header indent and the subsequent sibling indent collide. Example input `{"messages":[{"config":{"timeout":10},"role":"user"}]}`:
    ```
    Messages:
      - Config:
        Timeout:  10
        Role:  user
    ```
    `Role` is rendered at the same column as `Timeout` even though it's `Config`'s sibling, not its child. Root cause: when `firstFieldInline=true` and `i==0` is a header, the recursive `renderObject` for its children uses `depth+1` → `indent = "    "` (4 spaces). The `i>0` sibling uses `indent + markerSpaces` = `"  " + "  "` = `"    "` (also 4 spaces). The post-marker children need an extra indent unit so they sit at column 6 while siblings stay at column 4. The existing tests don't catch this — `ArrayOfObjects` test uses two scalar fields, `EmptyElement` uses a single field. Adding a `messages:[{nested:{x:1}, sibling:2}]` test would reproduce. Likely a few-line fix in `renderObject`'s recursive call site (pass an additional indent when first-field-as-header is detected) plus a sibling-indent adjustment, but the cleanest fix is probably to emit a newline after a first-field header in sequence-item mode and treat all subsequent fields as ordinary children at `depth+1` indent.

- *`internal/output/human_formatter.go:240-242` (depth-cap path) vs `:248-251` (below cap) — null at depth cap renders inconsistently.* Below the cap, `nil` renders as `(none)`. At `depth >= maxNestedDepth`, the early-return calls `compactJSON(value)` which returns `""` for nil, so a null leaf at the cap renders as a blank value (label, colon, nothing). Real-world impact is tiny (you need a chain of 5+ nesting levels ending in null), but the inconsistency is easy to remove with a `nil` short-circuit ahead of the depth check.

- *Lint claim in PR body vs CI.* PR body says `make lint` is broken on clean main (golangci-lint go1.23 vs go1.25 toolchain) and unrelated to this change. The PR's `lint` check is green on CI. Worth a one-line resolution in the description — either the runner uses a tolerant version, or the head fixed it incidentally, or the description is stale. Not blocking; just a "say what changed" hygiene point so future readers aren't confused.

*Nits:*

- `humanizeKey` uppercases only the first rune of each `_`/`-`/`.`-separated segment, so `amount_usd` → `Amount Usd` (and any acronym-like segment → title case). Pre-existing behavior, unchanged here, and the deep-nesting test bakes it in as expected output. Not a regression. Worth a future cleanup pass with an acronym allowlist.
- The README example block doesn't have a code-fence language identifier, which is fine in Slack but renders without highlighting on GitHub — pre-existing style in the file, leaving alone.

*Questions:*

- Were any production-shaped responses with array-of-objects-containing-nested-config actually tested against this? `heygen auth status --human` (the motivating example in the PR body) returns mostly flat scalars plus a 2-level `wallet.auto_reload`, which doesn't exercise the sequence-item-with-nested-first-field path. If there's an endpoint whose response shape is `{messages: [{tools: [...], role: "..."}]}` or similar, that's where the bug would surface to users first.

*What I didn't verify:*
- Wide-character / RTL / emoji label alignment — the diff uses `lipgloss.Width` consistently so it should be fine, but I didn't run the binary against a payload with CJK or RTL keys.
- Windows line-ending behavior on the multiline-string continuation path. CI shows `test (windows-latest)` green so probably fine.

*CI:* All 9 checks green (WIP, goreleaser-check, govulncheck, lint, pr-template, secrets, test on ubuntu / macos / windows).

_Review by Rames D Jusso_

somanshreddy:
Thanks for the careful review. All three concerns addressed in `2d9574c`:

**Concern 1 (array-of-objects, nested first field) — fixed.** You were exactly right about the root cause: a sequence item's body was rendered at `depth` with a `+markerSpaces` hack for siblings, so a nested first field's children recursed from that same `depth` and missed the `- ` marker's 2-column shift. Fix: render the element body at `depth+1` (the `- ` marker is exactly one indent unit wide, so it occupies precisely that level), drop the `+markerSpaces` for `i>0`, and set `markerPad=len(indent)` for `i==0`. Your example now renders:
```
Messages:
  - Config:
      Timeout:  10
    Role:  user
```
`Timeout` (Config's child) at col 6, `Role` (element sibling) aligned with `Config` at col 4. Added `TestHumanFormatter_Data_KeyValue_ArrayOfObjects_NestedFirstField` using your exact payload. Scalar-only sequence items are byte-identical to before (verified: the marker prefix width equals one indent unit, so the math collapses to the old values).

**Concern 2 (null at depth cap) — fixed.** Moved the `value == nil` check ahead of the `depth >= maxNestedDepth` guard, so a null leaf at/beyond the cap renders `(none)` instead of `compactJSON(nil)` -> blank. Added `TestHumanFormatter_Data_KeyValue_NullAtDepthCap` (6-level chain ending in null).

**Concern 3 (lint claim vs CI) — clarified in the PR body.** The `lint` CI check is green because CI runs a go1.25-compatible golangci-lint. The "broken" note was about the local devbox only, whose installed golangci-lint is built for go1.23 and refuses the go1.25 module. Environment issue, not code. Reworded so future readers aren't confused.

**Your question (production-shaped `{messages:[{nested..., role}]}`):** that shape is exactly what tripped the bug, and it wasn't previously exercised (the old tests used scalar-only and single-field elements). The new nested-first-field test covers it directly, so it won't regress.

**Nits — acknowledged, no change:** `humanizeKey`'s `Amount Usd` acronym handling is pre-existing; agreed it's worth a future acronym-allowlist pass. README code-fence language identifier left as the file's existing style.

Re-reviewed with Codex after the fixes (clean). Thanks again for catching the nested-first-field case.

somanshreddy:
Follow-up on the acronym nit: rather than defer it, I went ahead and addressed it in this PR (`23372c3`).

Added a small acronym allowlist to `humanizeKey` (`id`, `url`, `api`, `srt`, `vtt`, `tts`, `ssml`, `json`, `uuid`, `ai`, `hd`, `sd`, `fps`) so those segments render fully uppercase instead of title-cased. Since `humanizeKey` is the single chokepoint shared by the table auto-column headers and the key-value detail path, both surfaces improve; curated columns hard-code their own headers, so they're unaffected.

Verified on live data:
```
ID:                     a4ef5ef8160c413f857a8fa51fa3e141
Default Voice ID:       9e62cdbe830c4f82b53a6efdff3b0e86
Group ID:               1524842831f74dd593c6da4082f1ed88
Supported API Engines:  avatar_v, avatar_iv
Video URL:              https://...
```

Updated the affected test assertions (`Id` -> `ID`, `Voice Id` -> `Voice ID`) and noted the behavior in the README. The allowlist is intentionally conservative (exact split-segment match, lowercased lookup) to avoid false positives on normal words. So this thread is resolved within the PR rather than left as future work.

james-russo-rames-d-jusso (commented):
*Approve* — all three R1 items resolved at HEAD (`23372c3`), pinning tests landed for both code fixes, and Somansh proactively folded the `humanizeKey` acronym nit into the PR rather than deferring it. CI green across the board.

*R1 status:*

- *Concern 1 — sequence-item nested-first-field indent collision (`human_formatter.go:266-278`).* ✅ Resolved. `renderObjectArray` now emits the `- ` marker then recurses into `renderObject` at `depth+1` with `firstFieldInline=true`; `renderObject` uses `fieldIndent=""` + `markerPad=len(indent)` for `i==0` and the normal `fieldIndent=indent` for `i>0`. Because `len(seqMarker)==len(indentUnit)`, the marker prefix occupies exactly one indent unit, so the math collapses cleanly: nested children of the first field recurse at `depth+2` (column 6), sibling scalars sit at `depth+1` (column 4). Shape matches my suggested fix.
- *Pinning test for Concern 1.* ✅ `TestHumanFormatter_Data_KeyValue_ArrayOfObjects_NestedFirstField` (`human_formatter_test.go:501`) uses the exact `{"messages":[{"config":{"timeout":10},"role":"user"}]}` payload from my R1 and asserts the corrected output (Timeout at col 6, Role at col 4 aligned under Config). Will catch any regression of this specific shape.
- *Concern 2 — null at depth cap inconsistency (`:240-242` vs `:248-251`).* ✅ Resolved. `renderField` now short-circuits on `value == nil` ahead of the `depth >= maxNestedDepth` guard, so null leaves render `(none)` uniformly at every depth. `TestHumanFormatter_Data_KeyValue_NullAtDepthCap` (`:523`) pins it with a 6-level chain ending in null.
- *Concern 3 — lint claim vs CI in PR body.* ✅ Resolved. PR body reworded: "The `lint` CI check is green (CI runs a go1.25-compatible golangci-lint). `make lint` fails only on the local devbox, whose installed golangci-lint is built for go1.23..." — no future-reader confusion.
- *Nit (acronym handling in `humanizeKey`).* ✅ Resolved-differently — Somansh elected to address it in-PR (`23372c3`) rather than defer. Added a conservative acronym allowlist (`id`, `url`, `api`, `srt`, `vtt`, `tts`, `ssml`, `json`, `uuid`, `ai`, `hd`, `sd`, `fps`) keyed on exact lowercased segment match, so `video_url` → `Video URL`, `group_id` → `Group ID`, normal words like `amount_usd` still get title-cased (`usd` isn't in the allowlist — pragmatic stopping point). Updated test assertions and README accordingly.
- *Nit (README code-fence language identifier).* Acknowledged, no change — pre-existing file style. Fine.

*What changed at R2 (`7b69d07` → `23372c3`):*

- `2d9574c` — the two code fixes + their pinning tests.
- `bc5884a` — README note that long scalar values (pre-signed URLs etc.) may wrap in narrow terminals; points users at JSON+jq for clean extraction. Small docs nit, sensible addition given the wider columns the formatter now emits.
- `23372c3` — acronym allowlist on `humanizeKey` + test assertion updates + README mention.

*New findings:*

- *Acronym allowlist scope.* `humanizeKey` is shared between the table-auto-column header path and the new `--human` key-value path, so a script that *was* scraping table headers (e.g. parsing the `Id` column header) would now see `ID`. PR body already labels `--human` as may-change presentation; this isn't `--human` per se, but it's the same `humanizeKey` chokepoint. Verified by reading the diff that *curated* table columns hard-code their own headers (not via `humanizeKey`), so only auto-column headers are affected. Acceptable risk for the same opt-in surface the PR is reshaping; flagging for awareness, not blocking.
- *Allowlist is intentionally conservative.* I'd nudge `usd` / `eur` / `gbp` / `cad` (currency codes) as a future addition since `amount_usd` → `Amount Usd` is the canonical example from R1's nit, but Somansh's framing — only add segments that are unambiguously acronyms vs. real words — is a defensible cutoff. Leave alone.

*CI:* All 9 checks green at `23372c3` — `test (ubuntu/macos/windows-latest)`, `lint`, `pr-template`, `secrets`, `govulncheck`, `goreleaser-check`, `WIP`. Latest run 2026-06-14 04:24.

Net: clean fix, scope discipline (only one out-of-scope extra: the acronym pass, which is itself a tightening), good test coverage. Ready to merge.

_Review by Rames D Jusso_

jrusso1020 (approved):
Approved. All three R1 items resolved at `23372c3` with pinning tests for both code fixes; acronym allowlist nit folded in. All 9 CI checks green.

PR #181, opened 2026-06-22
feat(cli): hide under-development commands from help via a hard-coded list

somanshreddy:
## Summary
Adds a hard-coded `hiddenCommands` allowlist so a generated command can be omitted from `--help` listings while staying fully functional , for endpoints that exist in the OpenAPI spec but aren't announced yet (under development). This is the CLI analog of the API's `x-excluded` (reachable, not advertised).

## Motivation
The codegen is binary today: a command is either generated (visible in help + functional) or excluded (`x-cli-visible: false`, not generated at all). There was no "works if you know it, hidden from help" posture.

## How it works
- A hard-coded `hiddenCommands` map keyed by `METHOD /path` (stable across command renames) sets Cobra `Hidden`. The spec stays the source of truth for what exists; this list is the CLI-side decision of what is discoverable.
- A hidden command is omitted from help listings but runs if invoked, and its own `--help` works.
- `hideEmptyGroups` also hides an intermediate group left empty when its only leaf is hidden, so a multi-word command doesn't leak a phantom group entry into help.
- Hides `GET /v3/assets/search` (asset search) pre-launch.

## Testing
Unit tests for the allowlist match (case-insensitive method), the Cobra `Hidden` wiring, and `hideEmptyGroups`. Verified on a real build that `asset search` is absent from `heygen asset --help` for both the single-leaf and multi-word command shapes, while remaining reachable and runnable with a working `--help`.

### DIFF
--- a/cmd/heygen/builder.go
+++ b/cmd/heygen/builder.go
@@ -21,6 +21,25 @@ import (
 	"golang.org/x/term"
 )
 
+// hiddenCommands lists generated commands that remain fully functional but are omitted from
+// `--help` listings (Cobra Hidden) — for endpoints that exist in the OpenAPI spec but aren't
+// announced yet (still under development). The spec is the source of truth for what exists; this
+// hard-coded list is the CLI-side decision of what's discoverable. A hidden command still runs if
+// invoked directly, and its own `--help` works. Keyed by "METHOD /path" so it survives command
+// renames (e.g. an x-cli-action verb change).
+// To surface a command in --help, remove its entry here. This list is transitional, for endpoints
+// under development pre-launch; for a permanent hidden-from-help posture, plumb an `x-cli-hidden`
+// extension from the OpenAPI spec instead of hard-coding it here, so the spec stays the source of
+// truth.
+var hiddenCommands = map[string]bool{
+	"GET /v3/assets/search": true, // asset search — pre-launch, not yet announced
+}
+
+// isHiddenCommand reports whether the command for this spec should be omitted from help listings.
+func isHiddenCommand(spec *command.Spec) bool {
+	return hiddenCommands[strings.ToUpper(spec.Method)+" "+spec.Endpoint]
+}
+
 // buildCobraCommand creates a Cobra command from a command.Spec.
 // It registers typed flags, sets up positional arg validation, and wires
 // RunE to build an Invocation and execute it through the HTTP client.
@@ -37,6 +56,7 @@ func buildCobraCommand(spec *command.Spec, ctx *cmdContext) *cobra.Command {
 		Short:   spec.Summary,
 		Long:    description,
 		Example: strings.Join(spec.Examples, "\n"),
+		Hidden:  isHiddenCommand(spec),
 		Args: func(cmd *cobra.Command, args []string) error {
 			if isSchemaRequest(cmd) {
 				return nil
--- a/cmd/heygen/help_flatten.go
+++ b/cmd/heygen/help_flatten.go
@@ -176,3 +176,20 @@ func visibleChildren(cmd *cobra.Command) []*cobra.Command {
 func isLeafCommand(cmd *cobra.Command) bool {
 	return len(visibleChildren(cmd)) == 0
 }
+
+// hideEmptyGroups hides any group command left with no visible children — e.g. an intermediate
+// group whose only leaf is in hiddenCommands. Without this, such a group reads as a leaf to the
+// flattened help and renders as a phantom entry. Runs bottom-up so a group whose children all
+// become hidden is itself hidden. The underlying command stays reachable and runnable; only its
+// help listing is suppressed.
+func hideEmptyGroups(cmd *cobra.Command) {
+	for _, child := range cmd.Commands() {
+		hideEmptyGroups(child)
+	}
+	// A command with subcommands but no visible ones is an empty group (e.g. its only leaf is in
+	// hiddenCommands). Hide it. Leaves have no subcommands, so they're unaffected; intermediate
+	// groups carry a help-showing RunE, so we must not gate on Runnable().
+	if cmd.HasSubCommands() && len(visibleChildren(cmd)) == 0 {
+		cmd.Hidden = true
+	}
+}
--- a/cmd/heygen/hidden_commands_test.go
+++ b/cmd/heygen/hidden_commands_test.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/heygen-com/heygen-cli/internal/command"
+	"github.com/spf13/cobra"
+)
+
+func TestIsHiddenCommand(t *testing.T) {
+	hidden := &command.Spec{Group: "asset", Name: "search", Method: "GET", Endpoint: "/v3/assets/search"}
+	if !isHiddenCommand(hidden) {
+		t.Error("GET /v3/assets/search should be hidden")
+	}
+	// Method match is case-insensitive (specs may carry a lowercased method).
+	lower := &command.Spec{Group: "asset", Name: "search", Method: "get", Endpoint: "/v3/assets/search"}
+	if !isHiddenCommand(lower) {
+		t.Error("method match should be case-insensitive")
+	}
+	visible := &command.Spec{Group: "video", Name: "list", Method: "GET", Endpoint: "/v3/videos"}
+	if isHiddenCommand(visible) {
+		t.Error("GET /v3/videos should not be hidden")
+	}
+}
+
+func TestBuildCobraCommand_HiddenWiring(t *testing.T) {
+	ctx := testCmdContext()
+
+	hidden := buildCobraCommand(&command.Spec{
+		Group: "asset", Name: "search", Method: "GET", Endpoint: "/v3/assets/search",
+		Summary: "Search assets", Examples: []string{"heygen asset search --query x"},
+	}, ctx)
+	if !hidden.Hidden {
+		t.Error("asset search should be Cobra-Hidden (functional but omitted from help listings)")
+	}
+	// A hidden command must still be runnable and have its own help.
+	if hidden.RunE == nil {
+		t.Error("hidden command should still be runnable (RunE wired)")
+	}
+
+	visible := buildCobraCommand(&command.Spec{
+		Group: "video", Name: "list", Method: "GET", Endpoint: "/v3/videos",
+		Summary: "List videos", Examples: []string{"heygen video list"},
+	}, ctx)
+	if visible.Hidden {
+		t.Error("video list should not be hidden")
+	}
+}
+
+// TestHideEmptyGroups covers the multi-word case: when a hidden leaf's only sibling-set under an
+// intermediate group is itself, the intermediate group must also be hidden so it doesn't leak into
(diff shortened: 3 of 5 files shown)


james-russo-rames-d-jusso (commented):
**Reviewed at HEAD `3b53fee8`.** Looked at this paired with EF#40256 (the upstream `cli_action=True` + flag-taxonomy PR). The two are designed as a pair: EF establishes "still ships via codegen + announced via CLI"; this PR adds the missing "ships but stays off `--help`" posture. Design is coherent.

## Cross-repo verification table

The motivation here is filling a gap the existing `x-cli-visible` flag can't cover. Worth mapping the surface taxonomy across both repos:

| Posture | Spec emission (EF) | heygen-cli behavior |
|---------|-------------------|---------------------|
| Public (default) | (no extension) | Codegen, visible in `--help`, runnable |
| `cli_visible=False` | `x-cli-visible: false` | **Not codegen'd at all** — command doesn't exist (filtered at codegen/grouper.go:342) |
| `cli_visible=True` + `hiddenCommands` entry (this PR) | (no `x-cli-visible`) | Codegen'd, `Cobra.Hidden=true`, runnable but off `--help` |

The two posture differ exactly as Somansh's body claims: `cli_visible=False` = "doesn't exist if typed"; `hiddenCommands` entry = "works if you know it, hidden from discovery." This is the correct shape — `x-excluded` (docs) is the precedent for "reachable but not advertised," and this brings the CLI's posture set to parity with that.

## 🟢 Resolved

- **Key choice (`METHOD /path`) survives renames.** This is the right primary key, since EF#40256 specifically renames the command from `asset search list` → `asset search` via `x-cli-action`. The builder.go:36 lookup keys on `spec.Method + spec.Endpoint` (uppercase-normalized at :36), so the entry stays valid before and after the codegen sync runs.

- **`hideEmptyGroups` solves the multi-word leaf case.** Pre-codegen-sync, `gen/asset.go:172` has `Name: "search list"`, which `registerSpecCommand` (root.go:153-167) splits into `["search", "list"]` via `commandPathParts` — `search` becomes an intermediate group, `list` the leaf. Just hiding `list` would leave `search` rendering as a phantom entry to flattened help. The recursive `hideEmptyGroups` (help_flatten.go:185) walks bottom-up and hides empty intermediates. Tested at `TestHideEmptyGroups`. After codegen sync drops the `list` token, `asset search` becomes a single leaf directly under `asset`, and `hideEmptyGroups` becomes a no-op for this endpoint — but the helper is correct general infra for any future hidden-multi-word case.

- **Build order in `newRootCmd` is right** (root.go:50-55): `registerGroups` → `attachCustomCommands` → `attachDeprecatedAliases` → `hideEmptyGroups` → `installFlattenedHelp`. `hideEmptyGroups` runs AFTER all commands are attached and BEFORE flattened help captures the visible set. Same order in `newRootCmdWithSpecs` (:117-122). 

- **Case-insensitive method match** (builder.go:36, `strings.ToUpper`) — sensible defense, since spec sources can carry lowercased methods. Tested at `TestIsHiddenCommand` lower-case case.

- **Coexists cleanly with existing `Hidden` semantics** — `cmd/heygen/aliases.go:81-84` (`ensureHiddenGroup` + `leaf.Hidden = true`) already uses Cobra `Hidden` for deprecated-alias paths. The new `hideEmptyGroups` walk operates on the same field but at a different lifecycle moment (after all attachment) and on a different population (intermediate groups, not alias leaves). No interaction risk I can see.

## ⚠️ Concerns

- **Stale `Examples` strings post codegen-sync.** `gen/asset.go:181-182` (and `codegen/examples/asset.yaml:6,8`) say `heygen asset search list --query …`. After EF#40256 lands, codegen regenerates the Spec with `Name: "search"`, but `Examples` is sourced from the *human-maintained* `codegen/examples/asset.yaml` (strict copy at codegen/grouper.go:180) — `codegen/main.go:77-90` only validates presence, not content. So the example strings will stay stale and tell users to type `heygen asset search list …`, which won't parse anymore. This is a concrete user-facing footgun, but it lives in the codegen-sync follow-up, not this PR. Worth either: (a) calling out in this PR body that the sibling examples need updating, or (b) preempting by updating `codegen/examples/asset.yaml` in this PR. Cheap fix.

- **`asset search` "if you know it" UX** — the body says "still runnable if you know the command." Verified the mechanics: `Cobra.Hidden=true` suppresses listing but doesn't gate execution; `RunE` is still wired (TestBuildCobraCommand_HiddenWiring confirms). But a user typing `heygen asset search --help` will see normal help output (Cobra renders help for Hidden commands when invoked directly) — confirm this matches your intent. If you want `--help` to also be gated, that's a separate flag.

## 🟡 Questions

- **Unhide path.** The body doesn't document what someone does when `asset search` is ready to announce. Options:
  - (a) Remove from `hiddenCommands` here.
  - (b) Set `cli_visible=True` (already done) + remove from `hiddenCommands` (redundant given current default — `cli_visible=True` doesn't add to the list).
  - (c) Something else if the codegen pipeline ends up plumbing `cli_visible` through to `Hidden`.

  Adding a one-line comment to the `hiddenCommands` map declaration documenting the unhide ritual would prevent the next person from guessing. Even just "// Remove the entry to surface in `--help`." inline.

- **Could `hiddenCommands` ever drift from EF intent?** If EF later wants this to be a long-term posture (not just pre-launch), the right place to encode that is in the OpenAPI spec via a new `x-cli-hidden` extension, not the CLI repo's hard-coded list. Worth a comment near the map declaration saying "transitional — endpoints under development; for permanent hidden-from-help status, consider plumbing `x-cli-hidden` from the spec instead." Not blocking — the current shape is fine for pre-launch — but it keeps a future drift conversation on rails.

## 🟢 Nits

- The `hideEmptyGroups` comment at help_flatten.go:178-184 is great — calls out specifically that intermediate groups carry a help-showing `RunE` so we mustn't gate on `Runnable()`. That's a non-obvious Cobra detail; thanks for spelling it out.
- `hiddenCommands` map literal could grow into a slice of structs with a `reason` field if it ever exceeds ~5 entries (current single-entry shape is fine). The trailing comment `// asset search — pre-launch, not yet announced` is already doing the job for now.

## CI

All checks green (test ubuntu/macos/windows, lint, govulncheck, goreleaser-check, pr-template, secrets).

_Review by Rames D Jusso_

somanshreddy:
Thanks @james-russo-rames-d-jusso. Addressed:

**Stale examples:** updated `codegen/examples/asset.yaml` now (not deferring). The asset-search examples drop the `list` token (`heygen asset search --query ...`). Since `gen/` isn't regenerated in this PR, the committed `gen/asset.go` keeps `Name: "search list"` with matching `search list` examples (consistent today); the codegen sync that lands with EF#40256 regenerates both from the new spec + the updated yaml together, so the rename and the example fix apply atomically. No stale-`list` window.

**Unhide ritual + drift (Q):** added a comment on the `hiddenCommands` map declaration: remove the entry to surface in `--help`, and that this list is transitional (pre-launch). For a permanent hidden-from-help posture, plumb an `x-cli-hidden` from the OpenAPI spec rather than this CLI-side list, so the spec stays the source of truth.

**`--help` on the hidden command (concern):** intended. The requirement was "works, and its own `--help` works, if you know the command." `Cobra.Hidden` suppresses listing but not execution or `cmd --help`, which is exactly the posture. No gating of `--help` desired.

james-russo-rames-d-jusso (commented):
**R2 verification at head `4b75eed4`.** R1 was at `3b53fee8`; diff between them touches `cmd/heygen/builder.go` (4 lines added — `x-cli-hidden` future-drift paragraph) and `codegen/examples/asset.yaml` (verb drop in two example commands).

### R1 → R2 verdict transitions

**1. ⚠️ Stale codegen examples → ✅ Resolved (in the right place).**
`codegen/examples/asset.yaml:6,8` now read `heygen asset search --query ...` (no `list` verb). The `gen/asset.go:181-182` lines I flagged at R1 still show the old strings — but those are auto-generated artifacts downstream of EF's spec + this YAML, and they'll mechanically pick up both the YAML examples and the `x-cli-action`-driven `Name: "search"` rename on the next `codegen: resync gen/ from EF <sha>` PR (the canonical sync cadence — see commits `955b586`, `26565fc`, `e6c2716`). YAML was the correct file to fix; `gen/` is the wrong layer to hand-edit. Footgun-aware: between this PR landing and the next resync, the actual binary is still `heygen asset search list`. The path-stable `hiddenCommands` map (`builder.go:34` keyed by `"GET /v3/assets/search"`) means the hide survives the rename — confirmed at `cmd/heygen/hidden_commands_test.go:14-22`.

**2. ⚠️ `--help` semantics on hidden command → 🟢 Documented-as-known (already in R1 tree) + ✅ test-pinned.**
`cmd/heygen/builder.go:28` comment already said "A hidden command still runs if invoked directly, and its own `--help` works." `TestBuildCobraCommand_HiddenWiring` (`hidden_commands_test.go:27-49`) test-pins that `hidden.RunE != nil` so the runnability invariant has a test.

**3. 🟡 Unhide ritual not documented → 🟢 Resolved by code comment (new at R2).**
`cmd/heygen/builder.go:30-31` now reads: "To surface a command in --help, remove its entry here." Single-line answer to "what does someone do when this is ready to ship" — that's the right depth for an in-line maintenance instruction.

**4. 🟡 Possible `x-cli-hidden` future drift → 🟢 Resolved by code comment (new at R2).**
`cmd/heygen/builder.go:31-33` now reads: "This list is transitional, for endpoints under development pre-launch; for a permanent hidden-from-help posture, plumb an `x-cli-hidden` extension from the OpenAPI spec instead of hard-coding it here, so the spec stays the source of truth." Codifies the migration path; future drift has a name.

### Cross-repo coherence (re-verified)

- EF emits `x-cli-action: true` on `GET /v3/assets/search` (`external-api.json:11277`).
- heygen-cli `codegen/grouper.go:363` reads `Extensions["x-cli-action"]` and skips the terminal verb (`grouper_test.go:395-402` test-pinned).
- heygen-cli `hiddenCommands["GET /v3/assets/search"]` keyed by METHOD+PATH, survives the `search list` → `search` rename when the next codegen resync runs (`builder.go:29` documents this stability property).
- `TestHideEmptyGroups` (`hidden_commands_test.go:53-72`) pins the multi-word-group phantom-entry case Vance flagged.

All five legs verified at `4b75eed4`. ✅

### Net new findings
None. The two changes (doc-comment paragraph + YAML verb drop) are surgical and well-targeted.

### Verdict
LGTM at this HEAD. Leaving the stamp for Rames Jusso (the OG) per the lane split.

_R2 by Rames D Jusso_

jrusso1020 (approved):
Approving at `4b75eed4` — R2 verified clean by Rames D Jusso, all 4 R1 items resolved:

- Stale codegen examples — YAML updated; `gen/asset.go` is downstream auto-gen and follows automatically.
- `--help` semantics on hidden commands — already test-pinned + documented at R1.
- Unhide ritual + `x-cli-hidden` hardcoded-list drift — resolved with new code-comment paragraphs at R2 explaining the procedure.

The hardcoded-list shape is intentional pending the `x-cli-hidden` OpenAPI flag landing; once that ships, this gate can move to spec-driven via the cross-repo taxonomy work in EF#40256.

— Jerrai

PR #188, opened 2026-06-23
feat(cli): add --headers flag with allowlist validation

miguel-heygen:
## Summary

Adds a persistent `--headers` flag that sends extra HTTP headers on every API request. Headers are validated against an allowlist to prevent arbitrary header injection.

### Usage

```bash
heygen --headers "X-HeyGen-Client-Source: media-use" audio sounds list --query "upbeat tech"
```

### Validation

- **Allowlisted names only** — rejects headers not in the allowlist (currently: `X-HeyGen-Client-Source`)
- **Value validation** — must match `[a-zA-Z0-9_\-\.]+` (no spaces, special chars, or injection)
- **Format validation** — must be `Key:Value` (rejects malformed input)

```bash
# Allowed
heygen --headers "X-HeyGen-Client-Source: media-use" ...     # ✅

# Rejected
heygen --headers "Authorization: Bearer token" ...            # ❌ not in allowlist
heygen --headers "X-HeyGen-Client-Source: bad value!!" ...    # ❌ invalid characters
heygen --headers "garbage" ...                                # ❌ missing colon
```

### How it works

The flag is orthogonal to `X-HeyGen-Client-Origin` (auto-detected parent agent). Both coexist:
- `X-HeyGen-Client-Origin` → WHO spawned the CLI (e.g. `claude_code`, `cursor`)
- `X-HeyGen-Client-Source` (via `--headers`) → WHAT TOOL is calling (e.g. `media-use`)

To add new allowed headers, add an entry to `allowedHeaders` in `context.go`.

### Changes

| File | Change |
|------|--------|
| `root.go` | Register `--headers` as `PersistentFlags` `StringArray` |
| `context.go` | Parse `Key:Value`, validate name against allowlist, validate value charset |
| `client.go` | Add `extraHeaders` map, `WithExtraHeaders` option, set headers in `Do()` |

### Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] Valid header accepted, request succeeds
- [x] Non-allowlisted header name rejected with clear error
- [x] Invalid value characters rejected with clear error
- [x] Malformed format (no colon) rejected with clear error
- [x] Without `--headers`, no extra headers sent (backward compatible)

### DIFF
--- a/cmd/heygen/context.go
+++ b/cmd/heygen/context.go
@@ -5,8 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"net/textproto"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 
 	"github.com/heygen-com/heygen-cli/internal/auth"
 	"github.com/heygen-com/heygen-cli/internal/client"
@@ -123,10 +126,18 @@ func initContext(cmd *cobra.Command, version string, ctx *cmdContext) error {
 		return clierrors.NewUsage("HEYGEN_API_BASE uses HTTP which transmits API keys in plaintext. Set HEYGEN_ALLOW_HTTP=1 to allow.")
 	}
 
-	ctx.client = client.New(result.Key,
+	opts := []client.Option{
 		client.WithBaseURL(baseURL),
-		client.WithUserAgent("heygen-cli/"+version),
-	)
+		client.WithUserAgent("heygen-cli/" + version),
+	}
+	if hdrs, _ := cmd.Flags().GetStringArray("headers"); len(hdrs) > 0 {
+		parsed, err := parseAndValidateHeaders(hdrs)
+		if err != nil {
+			return err
+		}
+		opts = append(opts, client.WithExtraHeaders(parsed))
+	}
+	ctx.client = client.New(result.Key, opts...)
 
 	human, _ := cmd.Flags().GetBool("human")
 	if human {
@@ -134,4 +145,32 @@ func initContext(cmd *cobra.Command, version string, ctx *cmdContext) error {
 	}
 
 	return nil
+}
+
+// allowedHeaders is the set of header names that --headers accepts.
+// Add new entries here as new attribution or routing headers are defined.
+var allowedHeaders = map[string]bool{
+	"x-heygen-client-source": true,
+}
+
+var headerValueRe = regexp.MustCompile(`^[a-zA-Z0-9_\-\.]+$`)
+
+func parseAndValidateHeaders(raw []string) (map[string]string, error) {
+	parsed := make(map[string]string, len(raw))
+	for _, h := range raw {
+		k, v, ok := strings.Cut(h, ":")
+		if !ok {
+			return nil, clierrors.NewUsage("invalid --headers format: " + h + " (expected Key:Value)")
+		}
+		key := strings.ToLower(strings.TrimSpace(k))
+		val := strings.TrimSpace(v)
+		if !allowedHeaders[key] {
+			return nil, clierrors.NewUsage("header " + key + " is not in the allowlist")
+		}
+		if !headerValueRe.MatchString(val) {
+			return nil, clierrors.NewUsage("header value must be alphanumeric, hyphens, underscores, and dots only")
+		}
+		parsed[textproto.CanonicalMIMEHeaderKey(key)] = val
+	}
+	return parsed, nil
 }
--- a/cmd/heygen/headers_test.go
+++ b/cmd/heygen/headers_test.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"testing"
+)
+
+func TestParseAndValidateHeaders_Valid(t *testing.T) {
+	hdrs, err := parseAndValidateHeaders([]string{"X-HeyGen-Client-Source: media-use"})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if hdrs["X-Heygen-Client-Source"] != "media-use" {
+		t.Errorf("got %q, want %q", hdrs["X-Heygen-Client-Source"], "media-use")
+	}
+}
+
+func TestParseAndValidateHeaders_CaseNormalization(t *testing.T) {
+	hdrs, err := parseAndValidateHeaders([]string{"x-heygen-client-source: my-tool"})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if hdrs["X-Heygen-Client-Source"] != "my-tool" {
+		t.Errorf("got %q, want %q", hdrs["X-Heygen-Client-Source"], "my-tool")
+	}
+}
+
+func TestParseAndValidateHeaders_NotAllowlisted(t *testing.T) {
+	_, err := parseAndValidateHeaders([]string{"Authorization: Bearer token"})
+	if err == nil {
+		t.Fatal("expected error for non-allowlisted header")
+	}
+}
+
+func TestParseAndValidateHeaders_MalformedNoColon(t *testing.T) {
+	_, err := parseAndValidateHeaders([]string{"garbage"})
+	if err == nil {
+		t.Fatal("expected error for malformed header")
+	}
+}
+
+func TestParseAndValidateHeaders_InvalidValueChars(t *testing.T) {
+	_, err := parseAndValidateHeaders([]string{"X-HeyGen-Client-Source: bad value!!"})
+	if err == nil {
+		t.Fatal("expected error for invalid value chars")
+	}
+}
+
+func TestParseAndValidateHeaders_DuplicateCaseCollapse(t *testing.T) {
+	hdrs, err := parseAndValidateHeaders([]string{
+		"X-HeyGen-Client-Source: first",
+		"x-heygen-client-source: second",
+	})
+	if err != nil {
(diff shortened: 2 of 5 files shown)


somanshreddy (commented):
## Review

Adds a `--x-source` global flag that sends `X-HeyGen-Client-Source` on every request. The core mechanism is correct: persistent flag → read in `initContext` → client option → header in `Do()`, gated on non-empty so it stays backward compatible. The reformatted `"heygen-cli/" + version` line is a gofmt fix, no issue there.

Below combines a manual review with a Codex pass; both converged independently on the same three items. The naming one is worth settling before release since it touches both a public Go API and a user-facing flag.

### 1. Naming: follow the established `ClientOrigin` precedent

The codebase derives the Go identifier and flag from the **domain concept**, dropping the `X-` / `-HeyGen-` wire prefix:

| Option | Field | Header |
|---|---|---|
| `WithClientOrigin` | `clientOrigin` | `X-HeyGen-Client-Origin` |
| `WithXSource` (this PR) | `xSource` | `X-HeyGen-Client-Source` |

`WithXSource` / `xSource` / `--x-source` is the only identifier named after the HTTP `X-` prefix, and `--x-source` leaks header syntax into the user-facing surface (compare the clean sibling `--human`). Suggest aligning to the precedent:

- Option/field → `WithClientSource` / `clientSource`
- Flag → `--client-source`

```go
func WithClientSource(s string) Option {
    return func(c *Client) { c.clientSource = s }
}
```
```go
root.PersistentFlags().String("client-source", "", "Caller attribution tag sent as X-HeyGen-Client-Source header")
```
```go
if source, _ := cmd.Flags().GetString("client-source"); source != "" {
    opts = append(opts, client.WithClientSource(source))
}
```

### 2. Register the flag in both root builders

`newRootCmd` registers the flag, but `newRootCmdWithSpecs` (the spec-based builder used by `builder_test.go`) registers only `--human`. `initContext` is shared, so under the spec-based tree `GetString("...")` returns a silently-ignored error and no header is sent. It's safe today, but any builder-path test that exercises the flag would get a misleading pass (or `unknown flag`).

Better fix: extract the global-flag registration into one helper and call it from both builders. This also dedupes the existing `--human` line and keeps the two roots from drifting again.

```go
func addGlobalFlags(root *cobra.Command) {
    root.PersistentFlags().Bool("human", false, "Display output as a formatted table instead of JSON")
    root.PersistentFlags().String("client-source", "", "Caller attribution tag sent as X-HeyGen-Client-Source header")
}
```

### 3. Add tests for the new header path

`client_test.go` already tests the parallel header with both a set-case (`TestClient_Do_SetsClientOrigin`) and an empty-skip case (`TestClient_Do_OmitsClientOriginWhenEmpty`, which carries a comment on why empty must emit no header at all). This PR adds the same `if c.xSource != ""` guard but no test, so a future refactor dropping the guard would go uncaught. Mirror the existing pair:

- `TestClient_Do_SetsClientSource`
- `TestClient_Do_OmitsClientSourceWhenEmpty`
- plus one command-level test (`runCommand(t, srv.URL, "test-key", ..., "--client-source", "foo")`) asserting the httptest server receives `X-HeyGen-Client-Source: foo`, to prove the `initContext` wiring end to end.

### Minor: confirm the header name with the backend

There are now three overlapping source signals on the wire: `X-HeyGen-Source: cli` (the app), `X-HeyGen-Client-Origin` (who spawned the CLI), and the new `X-HeyGen-Client-Source` (what tool is calling). `X-HeyGen-Source` and `X-HeyGen-Client-Source` are easy to confuse. Worth confirming the backend actually keys on `X-HeyGen-Client-Source` under that exact name before merge.

somanshreddy (commented):
## Updated review (now `--headers` with allowlist validation)

The validation design is well-executed: allowlist on the header name, `[a-zA-Z0-9_\-\.]+` on the value, a `Key:Value` format check, all surfaced as `clierrors.NewUsage` (exit 2). This addresses the arbitrary-header concern from the thread. Below combines a manual review with a Codex pass; both converged on the same items. Three are blocking, one is parity cleanup.

### 1. (Blocking) No tests for a security-sensitive flag

`parseAndValidateHeaders` adds three error branches (missing colon, not-in-allowlist, bad charset) plus the happy path, and `WithExtraHeaders` adds a new send path. No test files are in the PR (3 changed, all non-test). `client_test.go` already has the precedent (`TestClient_Do_SetsClientOrigin` / `...OmitsClientOriginWhenEmpty`), and AGENTS.md requires asserting on exit codes + stderr envelope. Please cover:

- valid header → server receives `X-HeyGen-Client-Source`
- non-allowlisted name → exit 2
- malformed (no colon) → exit 2
- invalid value charset → exit 2
- mixed-case duplicate behavior (see #3)
- a client-layer test that `WithExtraHeaders` cannot override a reserved header (see #2)

### 2. (Blocking) `Do()` applies `extraHeaders` after the client's own headers, so user input can override reserved headers

```go
req.Header.Set("x-api-key", c.apiKey)
req.Header.Set("User-Agent", c.userAgent)
req.Header.Set("X-HeyGen-Source", "cli")
if c.clientOrigin != "" { req.Header.Set("X-HeyGen-Client-Origin", c.clientOrigin) }
for k, v := range c.extraHeaders {   // runs LAST → wins on collision
    req.Header.Set(k, v)
}
```

The allowlist (`x-heygen-client-source` only) prevents this today, but the client is the security sink and shouldn't depend solely on the CLI-layer allowlist. The day a colliding entry is added to `allowedHeaders`, user input silently clobbers `x-api-key` / `User-Agent` / `X-HeyGen-Source` / `X-HeyGen-Client-Origin` / `Content-Type`. Fix: set `extraHeaders` **first** (before the client's own `Set` calls) so the client's headers always win, and/or filter reserved names in the client layer. Add a test proving `WithExtraHeaders({"x-api-key": "bad"})` can't override the key.

### 3. (Blocking) Mixed-case duplicate headers produce a nondeterministic value

`parsed` is keyed by the trimmed *original* casing. `--headers "X-HeyGen-Client-Source: a" --headers "x-heygen-client-source: b"` yields two map entries that both canonicalize to the same wire header via `http.Header.Set`; ranging the map is nondeterministic, so the value that lands varies by run. Fix: normalize the key before storing (lowercase for map identity, optionally `textproto.CanonicalMIMEHeaderKey` for emission) so duplicates collapse deterministically (last-flag-wins) or are rejected.

### 4. (Non-blocking) `newRootCmdWithSpecs` doesn't register `--headers`

Only `newRootCmd` was updated; the spec-based test builder still registers just `--human`, so `GetStringArray("headers")` there returns a silently-discarded error and that harness can't exercise the flag. Extract an `addGlobalFlags(root)` helper called from both builders (also dedupes the existing `--human` line).

### Minor

- A new user-facing global flag plus the "add to `allowedHeaders` to extend" pattern is the kind of thing AGENTS.md says to capture in CONTRIBUTING.md; it's in the PR description but not in-repo docs.
- Cross-system note (not a code bug): the backend reads `X-HeyGen-Client-Source` zero times today and value-allowlists client attribution server-side, so this flag is a no-op until the companion backend change lands. Worth confirming that's tracked before merge.

somanshreddy (commented):
## Re-review — all prior blocking findings resolved ✅

Verified against the PR head in a worktree (build + tests, not just reading). `go build ./...` OK; `cmd/heygen` and `internal/client` tests pass.

1. **Tests added** — `headers_test.go` (valid, case-normalization, not-allowlisted, no-colon, bad charset, duplicate-collapse) plus `TestClient_Do_SetsExtraHeaders` and `TestClient_Do_ExtraHeadersCannotOverrideReserved`.
2. **Reserved-header ordering** — `Do()` now applies `extraHeaders` first, then the client's own headers override; covered by the override test.
3. **Casing nondeterminism** — key lowercased for the allowlist check and stored via `textproto.CanonicalMIMEHeaderKey`, so mixed-case duplicates collapse to one canonical key (deterministic last-wins).
4. **Root-builder parity** — extracted `addGlobalFlags(root)`, called from both builders.

Approving. The items below are all minor / non-blocking — fine to address here or in a follow-up.

### Minor

- **`internal/client/client.go`** — the new docstring says reserved headers "always win and cannot be overridden by user input," but `X-HeyGen-Client-Origin` is set conditionally (`if c.clientOrigin != ""`), so when origin is empty a caller using `WithExtraHeaders` could set it. Unreachable via the CLI (allowlist permits only `x-heygen-client-source`), but the library-level guarantee is slightly overstated. Either set the origin header unconditionally or soften the comment; the override test doesn't cover the origin case.
- **Test convention** — `headers_test.go` asserts error presence but not the exit code. AGENTS.md asks tests to assert on exit codes (these all go through `clierrors.NewUsage` → exit 2). A command-level `runCommand(... "--headers", "bad")` test asserting exit 2 + the stderr envelope would match the repo's bar.
- **Trivial (lint won't catch — no formatter enabled):** `net/textproto` import in `context.go` is out of alphabetical order (should precede `net/url`) and the test map literal is misaligned; `gofmt -w` cleans both.

### Cross-system note (not this PR)
The backend reads `X-HeyGen-Client-Source` zero times today and value-allowlists client attribution server-side, so the header is a no-op until the companion backend change lands. Worth confirming that work is tracked.

somanshreddy (approved):
Approving — all prior blocking findings (tests, reserved-header ordering, casing determinism, root-builder parity) are resolved and verified passing. Remaining notes are minor/non-blocking.

PR #191, opened 2026-06-24
auth: add internal OAuth + PKCE + loopback driver (PR 1 of OAuth stack)

jrusso1020:
## Description

PR 1 of 2 in the heygen-cli OAuth stack. **Foundation only — no
user-visible CLI surface changes yet.** This PR adds a self-contained
`internal/auth/oauth/` package that drives the HeyGen public OAuth
2.0 + PKCE flow. Nothing in the existing CLI, credential resolver,
or HTTP transport changes.

### What's in this PR

| File | Purpose |
|---|---|
| `oauth.go` | OAuth `Client` with `BuildAuthorizationURL`, `ExchangeAuthorizationCode`, `RefreshAccessToken`, `RevokeToken`. Endpoints + client_id are overridable so tests swap in `httptest.Server`. |
| `pkce.go` | RFC 7636 `GeneratePKCEPair()` (S256) + `GenerateState()`. |
| `loopback.go` | `StartLoopbackServer(ctx, redirectPath, expectedState)` — binds `127.0.0.1:0`, serves a single GET, constant-time state check, 5-min timeout, keep-alive-safe shutdown. |
| `browser.go` | `OpenBrowser(url)` — platform `open`/`xdg-open`/`rundll32` via `exec.Command`, with non-TTY + `BROWSER=none` + `HEYGEN_NO_BROWSER=1` headless fallbacks that print the URL. |
| `tokens.go` | `TokenResponse` struct + `Expired(now, skew)` helper. |

### Shared client_id with hyperframes-CLI

Both CLIs are public clients on the same HeyGen IdP and share the
baked-in `q2A2QRSke2LrFTPJhoDbHtXh` client_id; the HyperFrames CLI's
value is the authoritative one and must not drift. Endpoints follow
the same split: authorize on `app.heygen.com` (same origin as the
user's web session cookies), token + revoke on `api2.heygen.com`.

### What's deferred to PR 2

- `internal/auth/file_resolver.go:selectCredential` — the OAuth-aware
  credential resolver seam.
- `internal/client/client.go:Do` — `Authorization: Bearer` transport
  wiring.
- `cmd/heygen/auth_login.go` — new `heygen auth login` UX that calls
  this driver.

PR 2 ships those together so the `auth login` UX lands once with a
working end-to-end path.

### Notes

- No new dependencies — `golang.org/x/term` is already in `go.sum`
  from cross-platform stdin-isatty detection elsewhere. Platform
  browser launching uses `exec.Command`. No CGO.
- `internal/auth/credentials_file.go` already understands the JSON
  layout with an `oauth` block (mirrors hyperframes-CLI's shared file
  format), so the on-disk side is ready for PR 2.
- Headless / CI keeps using `HEYGEN_API_KEY` env (no file fallback in
  CI) per the OAuth-stack scope decision.

## Testing

- Table-driven unit tests for each file in the new package.
- `oauth_test.go` uses `httptest.NewServer` as a fake IdP and asserts
  happy-path + 400/401 (mapped to `ErrRejected`) + 5xx + context-cancel
  + bad-shape token responses (missing `access_token`, control chars,
  non-object).
- `loopback_test.go` covers happy path, state mismatch, IdP-error
  propagation, missing code, context cancel, non-GET rejection,
  argument validation.
- `pkce_test.go` verifies length, base64url shape, S256 round-trip,
  and uniqueness across calls.
- `browser_test.go` exercises the headless print path, opener-failure
  fallback, and the two env-var skips (`BROWSER=none`,
  `HEYGEN_NO_BROWSER=1`).
- `tokens_test.go` covers `Expired(now, skew)` edge cases including
  the proactive-refresh skew.
- All package tests pass with `-race -count=1`; full `make test`
  and `golangci-lint run ./...` clean locally.

### DIFF
--- a/internal/auth/oauth/browser.go
+++ b/internal/auth/oauth/browser.go
@@ -33,15 +33,24 @@ var printOut io.Writer = os.Stdout
 // genuine launch failures the caller may want to log; callers should
 // still proceed to wait on the loopback callback either way.
 //
-// Two env-var escape hatches mirror the hyperframes-CLI driver:
+// Env-var precedence on Linux/BSD:
 //
-//   - BROWSER=none     — never try to open a browser; print the URL.
-//   - HEYGEN_NO_BROWSER=1 — same, scoped to this CLI.
+//  1. HEYGEN_NO_BROWSER=1 — never try to open a browser; print the URL.
+//  2. BROWSER=none — same, mirrors the cross-distro convention.
+//  3. BROWSER=<command> (any other non-empty value) — invoke <command>
+//     with the URL as a single argument, per the de-facto Linux
+//     convention sketched in freedesktop.org and adopted by Python's
+//     webbrowser module, Git, etc.
+//  4. Fallback: xdg-open <url>.
+//
+// On macOS and Windows the platform-native opener is used regardless of
+// $BROWSER, since the conventional values there ("Safari", "Chrome",
+// "msedge") aren't shell commands.
 func OpenBrowser(url string) error {
 	if url == "" {
 		return errors.New("oauth: OpenBrowser called with empty URL")
 	}
-	if os.Getenv("BROWSER") == "none" || os.Getenv("HEYGEN_NO_BROWSER") == "1" {
+	if os.Getenv("HEYGEN_NO_BROWSER") == "1" || os.Getenv("BROWSER") == "none" {
 		printManualURL(url, "")
 		return nil
 	}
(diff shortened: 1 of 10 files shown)


jrusso1020 (commented):
## /code-review @ `f9337863` (PR 1: internal OAuth + PKCE + loopback driver)

**Verdict: LGTM-with-followups.** No blockers. Self-contained internal driver with zero live wiring (PR 2 does that), so the blast radius is effectively zero until the stack lands. PKCE generation, S256 hashing, base64url encoding, RFC 7636 verifier length, constant-time state comparison, header-injection scrubbing on tokens, and the revoke best-effort semantics are all correct. Test suite is thorough — happy path, IdP error, state mismatch, missing code, ctx cancel, non-GET, arg validation, 400/401/5xx/bad-shape on token endpoint. 10 findings (4 warn, 6 nit) below.

### ⚠️ Warnings (worth fixing before next push, non-blocking)

1. *Watchdog-goroutine leak on unexpected `Serve` error* — `loopback.go:95`. When `lb.server.Serve(listener)` returns a non-`ErrServerClosed` error (port-bind race, panic-recovered Serve failure), the goroutine delivers the error but never closes `lb.done` or invokes `shutdown()`. The 5-minute watchdog goroutine sits idle until the caller invokes the returned `cancel()`. Visible as zombie goroutines in long-running parent processes (test suites, daemons embedding the driver).

2. *`redirectPath` not validated for `?` / `#`* — `loopback.go:60`. A caller passing `redirectPath = "/oauth/callback?source=cli"` silently registers a dead route on the ServeMux (real callback hits `/oauth/callback?code=...&state=...` which doesn't match — the path-form fragment becomes part of the registered pattern). Loopback never delivers; user only learns via the 5-min timeout. Reproduced locally with httptest.

3. *`$BROWSER` env convention not honored on Linux* — `browser.go:44`. Driver only checks for the literal `"none"` and ignores the Linux convention where `$BROWSER` names the user's preferred command (`BROWSER=firefox`). Headless servers / minimal WMs that set `$BROWSER` still hit `xdg-open`, which may be missing or misconfigured. Worth checking whether hyperframes-CLI honors `$BROWSER` per the parity comment on line 36.

4. *Browser tests bleed external env* — `browser_test.go:33`. `TestOpenBrowser_TTYInvokesOpener` doesn't neutralize the developer's outer `BROWSER` / `HEYGEN_NO_BROWSER` env vars. Contributor with `BROWSER=none` in their shell sees the test fail because OpenBrowser short-circuits before invoking the test opener. Fix: `t.Setenv("BROWSER", "")` + `t.Setenv("HEYGEN_NO_BROWSER", "")` at the top of every browser test that asserts opener invocation.

### 🟡 Nits

5. *`numericField`'s `json.Number` branch is dead code* — `oauth.go:343`. Body decodes into `map[string]any` via `json.Unmarshal` without `Decoder.UseNumber()`, so all JSON numbers arrive as `float64`. Trim or switch to a `json.Decoder` with `UseNumber()` if precision-preservation on `expires_in` is a real concern.

6. *`joinPath` is dead production code* — `loopback.go:207`. Used only by `TestJoinPath_Helper`; comment claims "used by tests" but it lives in the production file. Move to `loopback_test.go` or inline the trivial concat in the test.

7. *Hand-rolled `htmlEscape` reimplements stdlib* — `loopback.go:235`. `html.EscapeString` is in stdlib, no new dep, fewer lines, faster, and matches Go's canonical entity table. The current implementation also allocates twice per rune.

8. *`tok.IssuedAt = c.Now()` stamped after parse* — `oauth.go:261`. For slow round trips the token looks younger than the server thinks, so `Expired()` refreshes later than necessary. Conservative direction (won't refresh too early), but stamping at request-send time would track server expiry more tightly. Edge case for short-lived tokens.

9. *`BuildAuthorizationURL` doesn't precondition-check args* — `oauth.go:117`. Empty `state`, `codeChallenge`, or `redirectURI` silently builds a malformed URL that the IdP 400s on. Defensive arg checks would surface the misuse at call site rather than via integration tests.

10. *`TokenError.Wrapped` is an exported mutable field on a public type* — `oauth.go:273`. The `Unwrap()` method already covers the public contract. Letting external callers do `te.Wrapped = nil` silently disables the `errors.Is(err, ErrRejected)` re-login UX cue. Prefer unexported `wrapped` + keep `Unwrap()` exported.

— Jerrai

somanshreddy:
Driver looks solid: PKCE/S256, constant-time state compare, loopback hardening (`?`/`#` rejection, keep-alive-safe shutdown), header-safe token validation, best-effort revoke. One tiny doc nit, non-blocking:

In `loopback.go`, `handleCallback`'s comment says non-matching requests get a "404 without revealing that a CLI is listening," but a non-GET on the matched path returns 405 (the `MethodNotAllowed` branch); only unknown paths get the mux 404. Worth tweaking the wording so it doesn't read as "always 404."

somanshreddy (approved):
Approving. Driver is clean: PKCE/S256, constant-time state compare, loopback hardening (?/# rejection, keep-alive-safe shutdown), header-safe token validation, best-effort revoke. The 404-vs-405 doc nit is addressed. Thanks!

PR #201, opened 2026-06-30
feat(cli): add feedback command

somanshreddy:
## Scope

**Surfaces:** CLI | **Module:** CLI / telemetry

## Summary

Adds a `heygen feedback` command so users (and especially agents) can report whether a flow worked or flag a bug, without leaving the terminal. Feedback is sent as an anonymous analytics event through the CLI's existing PostHog client, so there is no new backend and no API key required.

## How it works

The CLI already emits anonymous usage events (`COMMAND_RUN`, `COMMAND_RUN_COMPLETE`) through `internal/analytics`. This adds one more event in the same shape:

```
heygen feedback --rating <1-5> [--comment "..."]
```

- **`--rating`** is required, 1 to 5, anchored as 1 = broke/unusable, 3 = worked with friction, 5 = worked great. It maps to the `rating` property on a `CLI_FEEDBACK` event.
- **`--comment`** is optional free text, capped at 2000 characters (counted as runes, not bytes).
- The command needs **no auth** (`skipAuth`), so an agent can run it regardless of login state, and it makes **no API call**.
- It honors the existing telemetry opt-out: with analytics disabled it does not send, and says so in the JSON output rather than printing a false "thanks."

Think of it as the qualitative complement to the existing usage dashboard: command success rate tells you whether a command exited 0, feedback tells you whether it actually worked *well*.

## Design decisions

**Custom `CLI_FEEDBACK` event, not PostHog Surveys.** Surveys are built for in-app survey widgets and exist to feed PostHog's native survey UI; a CLI renders no widget, so the survey route would buy us nothing here (cross-tool aggregation was explicitly out of scope) while costing string-typed ratings and misleading semantics. A custom event matches the CLI's existing telemetry and keeps `rating` a real integer.

**Comment capped at 2000 characters, reject rather than truncate.** Unbounded free text flowing into a telemetry event invites pasted logs, stack traces, or secrets. 2000 runes is generous for a detailed bug note; longer input is rejected with a usage error (exit 2) so nothing is silently dropped. Detailed reports belong in a tracked issue, not this field.

**All CLI telemetry now sends `$ip: null`.** PostHog otherwise derives IP from the ingest request. Sending `$ip: null` makes the "anonymous" claim honest for both the new event and the existing usage events: PostHog neither stores nor geolocates the caller's IP.

**Bugs go to `feedback`, not auto-filed GitHub issues.** This is a public repo, and an agent auto-creating an issue could leak prompts, paths, or keys into a permanent public record. Docs steer agents to `feedback` (private, anonymous) and instruct them to surface tracking-worthy bugs to a human, who files the issue after reviewing it.

## Testing

Unit tests cover the analytics event (event name, `rating`/`comment` properties, empty-comment omission, disabled-returns-false, `$ip` null on all events) and the command (valid, rating-only, missing rating, out-of-range, comment over/at cap, and rune-vs-byte counting). All assert exit codes and the stderr error envelope. Reviewed to convergence with a Codex review loop (0 blocking).


miga-heygen (commented):
## Review: `feat(cli): add feedback command`

**SHA:** `1856dd9`
**CI:** All 8 checks pass (lint, govulncheck, secrets, goreleaser-check, tests on ubuntu/macos/windows).

---

### Verdict: LGTM — clean, well-designed addition

This is a tight PR. The design decisions are sound, the code follows existing patterns faithfully, and the test coverage is thorough. A few observations below, none blocking.

---

### What works well

1. **`$ip: null` in `baseProperties`** — excellent placement. Applying it to *all* events (not just feedback) retroactively fixes the IP-leakage gap for `COMMAND_RUN` and `COMMAND_RUN_COMPLETE` too. The test (`TestBaseProperties_SuppressesIP`) correctly validates both event types.

2. **Rune-counting for the comment cap** — `utf8.RuneCountInString` is the right call. The test (`TestFeedback_CommentCapCountsRunesNotBytes`) with 2000 `é` characters (4000 bytes, 2000 runes) proves the boundary. Reject-not-truncate is the right policy for a telemetry field.

3. **`skipAuth` annotation** — matches the established pattern (`newConfigCmd`, `newAuthCmd`). An agent can send feedback without credentials, as advertised.

4. **Omitting empty `comment`** — mirrors the `client_origin` pattern exactly. The test asserts absence (not empty string), which is correct for PostHog cohort hygiene.

5. **Test suite** — covers the happy path, rating-only, missing rating, out-of-range, comment-at-cap, comment-over-cap, rune-vs-byte, disabled client, and the analytics-layer properties. The `assertUsageEnvelope` helper validates the full JSON error shape, not just exit codes. Solid.

6. **Docs** — the README "Reporting bugs" section and CONTRIBUTING.md update are well-placed. The agent guidance ("do not open GitHub issues automatically") is a thoughtful guard for a public repo.

---

### Nits (non-blocking)

1. **Negative rating values** — `rating` is declared as `Int` on the cobra flag, so `--rating -1` is technically parseable. The validation `rating < 1 || rating > 5` catches it correctly (exit 2), so this is fine functionally. But the out-of-range test only covers `0`, `6`, `9` — a `-1` case would make the negative path explicitly documented in tests. Very minor.

2. **`command` property value** — `Feedback` calls `baseProperties("feedback")`, which sets `command: "feedback"`. The existing `CommandRun`/`CommandRunComplete` pass the full command path (e.g. `"heygen video list"`). For consistency, consider `"heygen feedback"` — this matters if anyone ever writes a PostHog query grouping by `command` across event types. Not blocking since `CLI_FEEDBACK` is a distinct event name and unlikely to be joined with `COMMAND_RUN` on the `command` dimension, but worth noting.

3. **`feedbackResponse.Message` when analytics is disabled** — the message includes `Unset HEYGEN_NO_ANALYTICS (or run 'heygen config set analytics true') to enable it.` The parenthetical is helpful, but note that the actual config key might be `analytics.enabled` or similar depending on the config layer. If the config command uses a different key name, the hint could confuse users. (I didn't trace the full config path, so this may already be correct.)

---

### Security

- No API key required or transmitted — feedback flows through the existing PostHog client with the hardcoded project key.
- Comment is length-bounded (2000 runes, rejected not truncated) — prevents log/stacktrace/secret paste from bloating telemetry.
- `$ip: null` suppresses IP storage in PostHog.
- No user-supplied input is used in any way that could cause injection (it's serialized to JSON and enqueued to PostHog's batch client).
- The "agents should not auto-file GitHub issues" guidance in docs is a good defense against PII/secret leakage into a public repo.

No security concerns.

---

Clean PR. Ship it.

— Miga

miguel-heygen (approved):
Reviewing at `1856dd9d`.

Specific checks that look solid:
- `cmd/heygen/feedback.go:20-65` keeps the command small and explicit: required 1-5 rating, rune-counted 2000-character comment cap, reject-not-truncate behavior, JSON response, and `skipAuth` so unauthenticated agents can report feedback.
- `cmd/heygen/root.go:47-50` and `cmd/heygen/root.go:127-130` register the command in both root builders, so production and spec-driven test construction stay aligned.
- `cmd/heygen/context.go:33-45` / `cmd/heygen/context.go:101-103` confirm the annotation actually bypasses credential resolution.
- `internal/analytics/analytics.go:91-106` emits a distinct `CLI_FEEDBACK` event and omits empty comments; `internal/analytics/analytics.go:113-123` sets `$ip: null` at the shared base-properties layer, so it fixes existing command events too.
- `cmd/heygen/feedback_test.go:177-256` and `internal/analytics/analytics_test.go:131-208` cover the important behavior: no-auth success, disabled analytics messaging, range/comment validation, rune-vs-byte boundary, feedback event properties, empty comment omission, and `$ip` suppression.

Additive to Miga's pass: I traced the disabled-analytics hint and `heygen config set analytics true` is the correct exposed key (`internal/config/config.go:30-36`, `cmd/heygen/config_cmd.go:57-60`). The negative-rating and `command: "feedback"` consistency notes are reasonable optional cleanups, not merge blockers.

All 8 checks are green.

**Verdict:** APPROVE
**Reasoning:** The new command follows existing CLI/auth/output patterns, has focused privacy guardrails, and the regression tests pin the main failure modes. No blocking correctness or security issues found.

— Magi

PR #205, opened 2026-07-07
cli: reclassify non-envelope errors and split 401/403 auth codes

somanshreddy:
## Scope

**Surfaces:** CLI (external v3 API) | **Module:** Error handling / observability

## Summary

Reclassifies CLI error responses so failures carry specific, actionable `error_code`s instead of a single opaque `error`, and so real v3 API errors are distinguishable from edge/infra failures. This is the first step toward driving the CLI's "server-side error rate" toward zero — most of that rate turned out to be mislabeling, not real failures.

## How it works

The CLI classifies every `≥400` response by whether it carries the v3 error envelope (`{"error":{"code","message"}}`):

- **Envelope present** (a `code` *or* `message`) → a real app error; preserve the app's own code. The old gate required a non-empty *message*, so a code-only envelope was dropped to generic `error`; that's relaxed.
- **No envelope** (gateway/WAF HTML, empty body, foreign JSON) → it did not come from the v3 app; classify by HTTP status.

Status → code (only when the code is absent or the non-specific literal `"error"`): 401→`unauthorized`, 403→`forbidden`, 400→`validation_error`, 402→`insufficient_credit`, 404→`not_found`, 409→`conflict`, 413→`payload_too_large`, 429→`rate_limited`, `5xx`→`unclassified_server_error`, unmapped `4xx`→`unclassified_client_error`. A non-empty message is always synthesized.

The observability payoff: `internal_error` (the app's own 5xx code, preserved) is now distinguishable from `unclassified_server_error` (a 5xx with no app envelope — edge, or the app crashing before its normal error path), and client-caused failures (out of credit, bad input, not permitted) carry their own codes instead of inflating one generic bucket.

## Design decisions

**401 and 403 are split.** They previously both became `auth_error`. 401 = not authenticated (re-login helps); 403 = authenticated but not permitted — plan, permission, or a flagged request — where re-login does *not* help. `enrichAuthHint` is gated to the not-authenticated codes so a 403 never gets a "log in" hint; the 403 family carries permission-oriented hints instead.

**Hints use the effective (status-derived) code,** so a code derived from status (e.g. a 402 with no code → `insufficient_credit`) still gets its hint. Consequence: a generic 404 now carries the standard `not_found` hint.

**No invented code for the no-envelope 5xx.** `unclassified_server_error` is the renamed existing (dormant) `server_error` fallback; `internal_error` stays the app's passthrough. The name deliberately implies neither transience nor a specific owner.

## Backward compatibility

Exit codes are unchanged **except one intentional correction**: a non-envelope 401/403 (previously `error` / exit 1) now exits 3, matching enveloped 401/403 which already exit 3. `error_code` values only sharpen where they were opaque; real v3 codes are unchanged. The `--help` footer and README exit-code table now read "auth / permission".

## Testing

Updated the behavior-change assertions and added coverage for: status-derived codes, the non-specific-code override, message synthesis, the non-login 403 hint (enveloped and non-envelope), non-envelope routing through `parseErrorResponse` (including the 401/403 exit-3 change), code-only envelopes, and the extended hint mappings. `make lint` clean; affected package tests green.

## Follow-up (not in this PR)

The dashboard server-list picks up `unclassified_server_error` at release (keeping `error`/`server_error` through the adoption tail). Part of a larger error-reduction effort; subsequent PRs cover response-parse/telemetry classification, the OAuth headless-login fix, client-side validation, and idempotency-key retry.


jrusso1020 (approved):
LGTM — approved. Verified the classification at source (`8c5a3c8c`):

- **Envelope detection is v3-shape-specific**: unmarshals into the nested `{"error":{code,message}}` struct and treats it as an app error only when `Code != "" || Message != ""`. Gateway HTML / empty / foreign JSON leave the nested `Error` zero-valued → correctly fall through to status classification. The code-OR-message relaxation is safe — it keys on the nested `error` object, so a foreign body with top-level `code`/`message` won't be misread as a v3 envelope.
- **Status→code override is guarded**: `isNonSpecific` (empty or literal `"error"`) resets to `""`, and each status case applies its code only `if code == ""` — real specific v3 codes are preserved, never clobbered.
- **401/403 split is correct**: 403 → `ExitAuth` + `forbiddenHint` ("re-authenticating will not help"), with a `hint == "" && statusCode == 403` catch-all so no 403 ever gets a "log in" hint; the login guidance (source-aware `enrichAuthHint`) is gated to `unauthorized` (401), which deliberately carries no static hint. 401→exit-3 confirmed via the passing `ExitAuth` test.
- **Message synthesis** never renders empty.
- Test coverage is thorough (status-derived codes, the non-specific override, message synthesis, the non-login 403 hint enveloped + non-envelope, non-envelope routing through `parseErrorResponse` incl. the exit-3 change, code-only envelopes, hint mappings).

Release-awareness (intentional + documented, not blocking): the non-envelope 401/403 exit-code change (1 → 3). It's a correctness improvement — makes non-envelope auth failures consistent with enveloped ones (already exit 3) and lets scripts *distinguish* auth failures rather than breaking a well-written one (exit 1 was the generic API/network bucket). README exit-code table + `--help` footer updated to match.

Minor (non-blocking): one commit carries a `Co-Authored-By: Claude` trailer — heygen-cli is public, so worth stripping on the squash if you keep OSS commit history clean.

CI green (9/9).

— Rames Jusso

PR #206, opened 2026-07-07
cli: surface error doc_url + param, fix rate-limit hint code

somanshreddy:
## Scope

**Surfaces:** CLI (external v3 API) | **Module:** Error handling

Stacked on #205 (error reclassification).

## Summary

Surfaces two fields the v3 API already provides but the CLI was dropping — `doc_url` and `param` — and fixes a rate-limit hint that never fired.

## How it works

- **`doc_url` + `param`:** the v3 error envelope carries `doc_url` (a per-code link to the error's documentation) and `param` (the request field that failed validation). The CLI parsed neither. This adds both to `CLIError` / `ToErrorEnvelope`, emitted only when present — so an agent or user gets the exact failing field and an authoritative doc link for every API error, with **zero hand-maintained mapping** (it scales automatically as the API adds codes).
- **Rate-limit hint fix:** the hint was keyed on `rate_limited`, but the v3 API emits `rate_limit_exceeded` for a 429 — so a real rate limit showed **no hint**. Now keyed on `rate_limit_exceeded` (and `quota_exceeded`), and the CLI's 429 empty-code fallback aligns to `rate_limit_exceeded` to match the API's name.

## Design decisions

**`doc_url` is pass-through only.** We surface the API's `doc_url` for enveloped errors; we do not fabricate URLs for CLI-derived / edge codes (they have no docs page) — those rely on the CLI hint.

## Testing

Unit tests: `doc_url`/`param` surfaced in the envelope and omitted when empty; `rate_limit_exceeded`/`quota_exceeded` hints present; 429 → `rate_limit_exceeded`. `make lint` clean; affected package tests green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Every command supports `--help`.
 | Aspect | Behavior |
 |--------|----------|
 | **stdout** | Always JSON. Even `video download` — binary writes to disk; stdout emits `{"asset", "message", "path"}` so you can chain on `.path`. |
-| **stderr** | Structured envelope on error: `{"error": {"code", "message", "hint"}}`. Stable `code` values for programmatic branching. |
+| **stderr** | Structured envelope on error: `{"error": {"code", "message", "hint", "param", "doc_url", "request_id"}}`. `code`/`message` are always present; `hint`/`param`/`doc_url`/`request_id` appear when applicable (`param`/`doc_url` are surfaced from the API for validation and documented errors). Stable `code` values for programmatic branching. |
 | **Exit codes** | `0` ok · `1` API or network · `2` usage · `3` auth / not permitted · `4` timeout under `--wait` (stdout contains partial resource for resume) |
 | **Request bodies** | Flags for simple inputs; `-d` for nested JSON (inline, file path, or `-` for stdin). Flags override matching fields. |
 | **Async jobs** | `--wait` blocks with exponential backoff; `--timeout` sets max (default 20m). 429s and 5xx retry automatically. |
@@ -172,7 +172,7 @@ Every command supports `--help`.
 Example error envelope:
 
 ```json
-{"error": {"code": "not_found", "message": "Video not found", "hint": "Check ID with: heygen video list"}}
+{"error": {"code": "not_found", "message": "Video not found", "hint": "Check ID with: heygen video list", "doc_url": "https://developers.heygen.com/docs/error-codes#not-found"}}
 ```
 
 ## Configuration
--- a/SKILL.md
+++ b/SKILL.md
@@ -69,7 +69,7 @@ heygen video get --response-schema
 ## Output Contract
 
 - **stdout**: JSON (always). This is the only output agents should consume.
-- **stderr**: JSON error envelope on failure: `{"error":{"code":"...","message":"...","hint":"..."}}`
+- **stderr**: JSON error envelope on failure: `{"error":{"code":"...","message":"...","hint":"...","doc_url":"...","param":"..."}}` (`hint`/`doc_url`/`param`/`request_id` present when applicable)
 - Do not pass `--human`. It produces unstructured text that cannot be parsed.
 
 ## Notes
--- a/internal/client/executor_test.go
+++ b/internal/client/executor_test.go
@@ -257,7 +257,7 @@ func TestExecute_RetryOn429(t *testing.T) {
 		calls++
 		if calls == 1 {
 			w.WriteHeader(http.StatusTooManyRequests)
-			_, _ = w.Write([]byte(`{"error":{"code":"rate_limited","message":"too many requests"}}`))
+			_, _ = w.Write([]byte(`{"error":{"code":"rate_limit_exceeded","message":"too many requests"}}`))
 			return
 		}
 		w.WriteHeader(http.StatusOK)
@@ -292,7 +292,7 @@ func TestExecute_RetryPreservesBody(t *testing.T) {
 		calls++
 		if calls == 1 {
 			w.WriteHeader(http.StatusTooManyRequests)
-			_, _ = w.Write([]byte(`{"error":{"code":"rate_limited","message":"too many requests"}}`))
+			_, _ = w.Write([]byte(`{"error":{"code":"rate_limit_exceeded","message":"too many requests"}}`))
 			return
 		}
 		w.WriteHeader(http.StatusOK)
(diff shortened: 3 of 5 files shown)


james-russo-rames-d-jusso (commented):
_Reviewed at e15db1eb2e2abba18001a6e9b595cb596cc03048._

Read the whole diff plus the surrounding `errors.go` / `APIError` / call-sites. Small, clean fix — the diagnosis matches the code, the fields plumb through end-to-end (`APIError.Param/DocURL` → `FromAPIError` → `CLIError` → `ToErrorEnvelope`), the omit-when-empty semantics work, and README/SKILL.md docs are updated. LGTM from my side; a couple of small observations below.

## Concerns

None blocking.

## Nits

- **`internal/client/executor_test.go:260,295`** — the 429-retry mocks still hardcode `{"error":{"code":"rate_limited",...}}`. These tests don't assert on the code content (they only verify retry behavior kicks in on a 429 status), so they still pass, but the fixture is now inconsistent with the PR body's assertion that "the v3 API emits `rate_limit_exceeded` for a 429." Worth a one-line sweep to match the new canonical name — the fixtures then double as documentation of what a real 429 body looks like. (nit)

- **`quota_exceeded` addition (`errors.go:215`)** — this code is added to the hint switch but never produced anywhere in the CLI itself (I grepped). If it's a v3 API code that can arrive on the wire, great — the hint pre-registration is correct. If it was speculative, no harm, but worth confirming against the API's canonical code list so the hint is actually reachable. (nit)

## Questions

- On the design decision that `doc_url` is pass-through only from the API — is any consumer of the CLI's stderr envelope (agent framework, doc generator, etc.) expected to open or render the URL? Right now it just serializes into JSON on stderr, so a hostile / drifted-API URL is a very small phishing surface (an LLM agent might follow `javascript:` or a private-network URL in an error). The pass-through-only stance is fine given the trust boundary is HeyGen's own API; flagging it here just so it's on the record if a future consumer starts opening these URLs.

## What I didn't verify

- Whether the v3 API code registry ever emits `rate_limited` on any code path — the fix assumes only `rate_limit_exceeded`. If any legacy edge/gateway still emits `rate_limited`, the hint would now be missing for that specific case (the empty-code fallback still catches it correctly).
- End-to-end: I didn't run the CLI against a real 429 to confirm the hint renders.

somanshreddy:
Thanks for the review. Addressed:

**Nit 1 — `rate_limited` 429 fixtures (`executor_test.go:260,295`):** fixed. Both mocks now emit `{"error":{"code":"rate_limit_exceeded",...}}`, so the fixtures match the canonical v3 429 code and double as documentation of a real 429 body. (Branch restacked, so SHAs updated and CI is re-running.)

**Nit 2 — `quota_exceeded` (`errors.go:215`):** confirmed reachable, not speculative. It's a canonical v3 code (`_STANDARD_ERROR_DEFINITIONS`, HTTP 429), so it can arrive on the wire and the pre-registered hint is correct. Kept.

**Question — `doc_url` pass-through as a phishing surface:** agreed the pass-through-only stance is fine given the trust boundary is HeyGen's own API (the CLI only talks to the HeyGen API base, and `doc_url` is first-party from the error envelope). Nothing in the CLI opens it today; it only serializes into stderr JSON. Noting for the record: if a future consumer ever auto-opens these, the cheapest guard with zero drift risk is a scheme check (surface only `https://`), which neutralizes `javascript:`/`data:`/`file:` without allowlisting a host that could move. Happy to add that if we ever want it, leaving it out for now.

**On "what I didn't verify" (does v3 ever emit `rate_limited`):** it doesn't. `rate_limited` is not in the API's `_STANDARD_ERROR_DEFINITIONS`; the API emits `rate_limit_exceeded` (429) and `quota_exceeded` (429), never `rate_limited`. So no code path loses its hint. The empty-code fallback also now synthesizes `rate_limit_exceeded`, matching the API.

james-russo-rames-d-jusso (commented):
_R2 — reviewed at 844c843._

Round 1 items resolved.

## Resolution

- **Nit 1** — stale `rate_limited` fixtures at `internal/client/executor_test.go:260,295`: ✅ addressed at 844c843. Both mocks now emit `"code":"rate_limit_exceeded"`, matching the canonical v3 429 code.
- **Q — `quota_exceeded` reachability**: ✅ confirmed. Somansh's comment cites `_STANDARD_ERROR_DEFINITIONS` (v3 API can emit it at 429). Pre-registered hint is correct.
- **Bonus — `doc_url` phishing surface**: acknowledged. Pass-through-only is the right stance given the CLI only speaks to the HeyGen API base; the scheme-check guard is noted for the day a consumer starts auto-opening. Right level of forward-planning without over-building.

`govulncheck` still red — same pre-existing `golang.org/x/crypto@v0.52.0` reachable via `selfupdate.init`, not a regression from this PR.

LGTM from my side. Leaving as a comment.

james-russo-rames-d-jusso (commented):
_R3 — reviewed at 844c843._

No new commits since R2. Still LGTM from my side — fixture rename (`rate_limited` → `rate_limit_exceeded`) is intact at `internal/client/executor_test.go:260,295`, `quota_exceeded` still in the hint switch, and the added `TestParseErrorResponse_SurfacesParamAndDocURL` integration test is a nice touch for the `param`/`doc_url` plumbing.

CI is unchanged too — `govulncheck` red is the same pre-existing `x/crypto@v0.52.0`-via-`selfupdate.init` (plus a fresh Go-stdlib TLS advisory `GO-2026-5856` in `crypto/tls@go1.25.11` that lands on `oauth/`, `output/`, and `retry.go` paths this PR doesn't touch — environmental, not caused by this PR).

Ready to ship from where I sit; leaving as a comment (approval routing lives elsewhere).

jrusso1020 (approved):
Approving at `844c843`. Verified: `Param`/`DocURL` plumb end-to-end (nil-guarded from `apiErr`, conditionally emitted in `ToErrorEnvelope`), the 429 `rate_limited`→`rate_limit_exceeded` fix with `quota_exceeded` sharing the retry hint, docs (README/SKILL) updated to match, and both integration tests (`SurfacesParamAndDocURL`, `SurfacesDocURLAndParam`) cover the JSON-parse and struct paths. All required checks green (test ×3 / lint / secrets / goreleaser). The `govulncheck` red is non-required and pre-existing (x/crypto via `selfupdate.init`), untouched by this PR. LGTM.

PR #228, opened 2026-07-21
fix(client): per-operation HTTP request timeouts

somanshreddy:
## Scope

**Surfaces:** API (CLI) | **Module:** HTTP client / timeouts

## Summary

The CLI applied one hard-coded 30s HTTP timeout to every request. That truncated roughly 1% of otherwise-successful video-create and asset-upload calls (whose server-side latency sits at or above 30s), while being needlessly loose for sub-second reads. This gives each operation class a timeout budget matched to its real latency.

## Root cause

A single `http.Client{Timeout: 30s}` was shared across all operations. But server-side latency differs by an order of magnitude by operation (Datadog, `api_service`, 30d):

- `POST /v3/videos` (create): p50 3.9s, **p99 30.8s** — create does synchronous work (quota/asset validation, analytics, DB insert), it is not a fast enqueue; already emitting some 504s.
- `POST /v3/assets` (upload): **p99 32.7s** — bounded by file size and client bandwidth, not server compute.
- Reads / lists: p99 < 2s.

So one global value is simultaneously too tight (create/upload) and too loose (reads).

## Fix

`timeoutForSpec` classifies each generated command by operation and picks a budget:

- multipart uploads → **10m** (matches the existing `video download` client)
- create/enqueue endpoints (those with a poll config) → **120s** (clears the ~30s p99 with headroom)
- everything else → **30s**

The policy is a runtime lookup in `cmd/heygen` (like poll configs and `--human` columns), not on the immutable generated `Spec`. It is applied via `client.SetTimeout` once per invocation in the builder (each CLI process runs a single command), covering both the plain and `--wait` paths. Deliberately not user-configurable yet — an override can be added when a real case needs one. No change to JSON output or any success path.

## Testing

Unit tests for `timeoutForSpec` (upload / create / read classes) and `client.SetTimeout`. Affected `go test` (`internal/client`, `cmd/heygen`) and `make lint` pass.

### DIFF
--- a/cmd/heygen/builder.go
+++ b/cmd/heygen/builder.go
@@ -111,6 +111,12 @@ func buildCobraCommand(spec *command.Spec, ctx *cmdContext) *cobra.Command {
 				}
 			}
 
+			// Per-operation HTTP timeout: uploads and create/enqueue calls run
+			// far longer than reads, so give each operation its own budget
+			// rather than one global value. Set once here (each invocation runs
+			// a single command); covers both the --wait and plain paths below.
+			ctx.client.SetTimeout(timeoutForSpec(spec))
+
 			// Poll config is looked up at call time, not attached to Spec.
 			// Spec is immutable (generated); poll configs are hand-written in cmd/.
 			pc := pollConfigs[spec.Group+"/"+spec.Name]
--- a/cmd/heygen/timeouts.go
+++ b/cmd/heygen/timeouts.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"net/http"
 	"time"
 
 	"github.com/heygen-com/heygen-cli/internal/client"
@@ -9,31 +10,39 @@ import (
 
 // Per-operation HTTP request timeouts. A single global timeout is wrong here:
 // server-side latency differs by an order of magnitude across operation classes
-// (reads p99 < 2s; create endpoints p99 ~30s because they do synchronous work;
-// uploads are bounded by file size and client bandwidth, not server compute).
-// These are deliberately not user-configurable yet — add an override only when
-// a real case needs one.
+// (reads p99 < 2s; writes run to a p99 of ~30s because they do synchronous work
+// and some enqueue async jobs; uploads are bounded by file size and client
+// bandwidth, not server compute). Deliberately not user-configurable yet — add
+// an override only when a real case needs one.
 const (
-	// timeoutRead covers reads/lists and quick POSTs. Matches the client default.
+	// timeoutRead covers GET reads and lists (server p99 < 2s). Tracks the
+	// client's built-in default so there is a single source of truth for 30s.
 	timeoutRead = client.DefaultTimeout
-	// timeoutCreate covers create/enqueue endpoints, which run to a p99 of ~30s
-	// server-side; 120s clears that with headroom before giving up.
-	timeoutCreate = 120 * time.Second
-	// timeoutUpload covers multipart uploads, gated by file size and bandwidth.
-	// Matches the dedicated video-download client budget.
+	// timeoutWrite covers every mutating call (POST/PUT/PATCH/DELETE): creates
+	// that do synchronous work or enqueue async jobs, plus other writes. Sized
+	// off the slowest observed case — POST /v3/videos runs to a Datadog p99 of
+	// ~31s (30d window) — with roughly 4x headroom before giving up.
+	timeoutWrite = 120 * time.Second
+	// timeoutUpload covers multipart uploads, bounded by file size and client
+	// bandwidth. Matches the dedicated video-download client budget.
 	timeoutUpload = 10 * time.Minute
 )
 
 // timeoutForSpec picks the per-request HTTP timeout for a generated command by
-// operation class. A runtime enhancement kept out of the immutable generated
-// Spec, like poll configs and --human columns.
+// operation class, keyed on the spec's HTTP method and body encoding — signals
+// that live on the immutable generated Spec. Method-driven, not coupled to the
+// hand-written poll-config set (which covers only 4 of the async-create
+// endpoints), so the classification can't silently drift as commands are added.
+// A runtime enhancement, kept out of the Spec like poll configs and --human
+// columns.
 func timeoutForSpec(spec *command.Spec) time.Duration {
 	if spec.BodyEncoding == "multipart" {
 		return timeoutUpload
 	}
-	// Create/enqueue endpoints are exactly the ones with a poll config.
-	if pollConfigs[spec.Group+"/"+spec.Name] != nil {
-		return timeoutCreate
+	if spec.Method == http.MethodGet {
+		return timeoutRead
 	}
-	return timeoutRead
+	// Every non-GET, non-upload call is a write (create / update / delete).
+	// Creates in particular are slow-synchronous or async-submit server-side.
+	return timeoutWrite
 }
--- a/cmd/heygen/timeouts.go
+++ b/cmd/heygen/timeouts.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"time"
+
+	"github.com/heygen-com/heygen-cli/internal/client"
+	"github.com/heygen-com/heygen-cli/internal/command"
+)
+
+// Per-operation HTTP request timeouts. A single global timeout is wrong here:
+// server-side latency differs by an order of magnitude across operation classes
+// (reads p99 < 2s; create endpoints p99 ~30s because they do synchronous work;
+// uploads are bounded by file size and client bandwidth, not server compute).
+// These are deliberately not user-configurable yet — add an override only when
+// a real case needs one.
+const (
+	// timeoutRead covers reads/lists and quick POSTs. Matches the client default.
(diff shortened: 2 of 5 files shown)


somanshreddy (commented):
Reviewed the per-operation timeout change — the core approach is right and well-tested (upload / create / read budgets, kept out of the immutable generated `Spec` like poll configs, `SetTimeout` documented as safe pre-request since `net/http` applies `Timeout` per-request). LGTM.

One low-severity observation (nonblocking):

**`timeoutForSpec` infers the "create" class from `pollConfigs[...] != nil`.** With `DefaultTimeout = 30s` and only 4 poll-config entries, every non-multipart operation that lacks a poll config falls to the 30s read budget. The comment asserts *"create/enqueue endpoints are exactly the ones with a poll config"* — that invariant is load-bearing here. If any create/enqueue (or otherwise slow-but-synchronous) endpoint doesn't have a poll config today, or one is added later without one, it silently gets the 30s read budget and can time out prematurely — the exact class this PR is trying to protect. Consider either deriving the class from the spec's method/semantics, or adding a test that walks all generated specs and asserts every create/enqueue op maps to `timeoutCreate`, so the coupling to `pollConfigs` can't drift unnoticed.

— Review by Somu

james-russo-rames-d-jusso (commented):
Reviewing `858c099` — per-op HTTP timeouts (`fix(client): per-operation HTTP timeouts (upload 10m, create 120s, read 30s)`).

Scope: adds `timeoutForSpec` (`cmd/heygen/timeouts.go`) + `Client.SetTimeout` (`internal/client/client.go:530`) and calls the pair once per invocation in `buildCobraCommand` (`cmd/heygen/builder.go:118`). Classification is spec-driven, not method/URL-driven — `BodyEncoding == "multipart"` → 10m, poll-config presence → 120s, everything else → 30s. Pairs with #229 which unifies how a tripped timeout is *classified* downstream on both the plain and `--wait` paths; the mechanics of "which budget applies" live here.

Verified against the tree at HEAD: `pollConfigs` in `cmd/heygen/poll_configs.go:7` covers exactly `video/create`, `video-translate/create`, `video-agent/create`, `lipsync/create`. The "matches the existing `video download` client" claim in the PR body checks out — `cmd/heygen/video_download.go:19` sets `downloadClient = &http.Client{Timeout: 10 * time.Minute}`. `client.DefaultTimeout` is `30 * time.Second` (`internal/client/client.go:21`), so `timeoutRead` genuinely tracks the prior constructor default.

### Blockers

None — the invariants the design leans on (one-command-per-process; `pollConfigs` == the create/enqueue set we care about; spec-level `BodyEncoding` is the reliable multipart signal) all hold in-tree.

### Concerns

- 🟠 **Classifier misses async creates that lack a poll config.** `pollConfigs` covers only `video/create`, `video-translate/create`, `video-agent/create`, `lipsync/create`. But several other create endpoints are explicitly described as asynchronous "submit → poll" in their generated specs and get `timeoutRead=30s`:
  - `gen/background-removal.go` — *"Submit a video to remove its background. Returns the created job; runs asynchronously, so poll Get Background Removal…"*
  - `gen/ai-clipping.go` — *"Submit a source video and return a job id immediately. The job runs asynchronously…"*
  - `gen/avatar.go` (`create`) — *"Avatar training is asynchronous."*
  - `gen/voice.go` (`clone create`) — *"Returns a voice_clone_id that can be polled via GET /v3/voices/{voice_clone_id}."*

  If any of these has a server-side p99 similar to `video/create`'s 30.8s, they'll exhibit the same ~1% truncation this PR set out to eliminate — just on different endpoints. Two paths forward:
  1. Move classification off "has a hand-written poll config in `cmd/`" onto a spec-level signal (e.g. an `Async` boolean codegen'd from an OpenAPI `x-async: true` extension, or a shared `enqueue` tag). More robust; declarative on the immutable spec, matching the rationale already stated in `timeouts.go` for keeping this out of `Spec`.
  2. Cheaper: add poll configs for the four endpoints above so they inherit `timeoutCreate` even if we're not ready to wire full `--wait` support for each. `timeoutForSpec` already treats poll-config presence as the sole signal, so this is a low-cost fix.

  Non-blocking because a follow-up can widen the budget when telemetry surfaces a bite, but calling out here so it's tracked.

- 🟠 **Callers outside `buildCobraCommand` silently inherit the 30s default.** `cmd/heygen/auth_status.go:28` and `cmd/heygen/video_download.go:87` both call `ctx.client.Execute(spec, inv)` with hand-built specs and never touch `SetTimeout`. Today both happen to be reads (`/v3/users/me`, `/v3/videos/{video_id}` GET), so 30s is correct — a coincidence, not an invariant. Pit-of-non-success for any future hand-written command that does a create or upload without going through `buildCobraCommand`. Consider one of:
  1. Push the budget selection into `Execute` itself, taking it from the spec (would break the "generated `Spec` is immutable" boundary, so probably not the right move).
  2. Add a comment on `Execute` documenting that callers outside the builder are responsible for `SetTimeout`, or wrap `Execute` in a helper that both sets the timeout and calls it.
  3. At minimum, a test-file assertion listing the currently-known non-builder callers so a new one has to update the list.

  Not blocking — current callers work — but this footgun grows in weight as more hand-written commands land.

- 🟠 **120s ceiling rationale isn't pinned in-tree.** p99=30.8s → 120s is ~4x headroom, which is defensible, but the number will sit here for a while. Was 120s picked off Datadog p99.9? A `p99 * K` heuristic? Server-team guidance? The comment on `timeoutCreate` says "clears that with headroom before giving up" — a one-line "picked because p99.9 in the 30d window is ~Xs" note (either in the const doc or the PR body pinned to the merge commit) would let a future engineer know what to re-check when the number needs to move. Non-blocking.

- 🟠 **No observability on the applied budget when a timeout fires.** #229's error message is `"request timed out: {URL error}"` and its hint doesn't cite the budget or class. A user hitting a truncation on `video create` sees the same error text as a user hitting it on `video list`. `timeoutForSpec` already knows the class + duration; threading that into the timeout error (either as a distinct hint or a `budget: "120s (create/enqueue)"` structured field on the error envelope) would meaningfully improve support triage. This is more of a #228+#229 stacked concern than pure #228 — flagging here because #228 is where the class information originates. Non-blocking.

### Nits

- 🟡 **No integration test covers the wiring.** `TestTimeoutForSpec` tests the classifier, `TestSetTimeout` tests the setter, but nothing asserts `buildCobraCommand` actually calls `ctx.client.SetTimeout(timeoutForSpec(spec))` per op class. A small table test that constructs a Cobra command for each of {multipart / pollable create / plain read} and inspects `c.httpClient.Timeout` after `RunE` would close the loop and lock the invariant.
- 🟡 **`timeouts.go:19` comment** says "Matches the client default" — worth being explicit ("`client.DefaultTimeout = 30s`") so a reader landing here without cross-file context sees the intent.

### Questions

- Are there any endpoints that legitimately need >30s and are NOT multipart uploads or pollable creates? SSE / streaming / long-polling shapes? If any exist today, they land in `timeoutRead` and will truncate. Quick spec-grep for `text/event-stream` or long-lived GETs would confirm.
- Is there an env-var escape hatch pending for enterprise users with unusual network topologies (behind slow proxies etc.)? The comment says "deliberately not user-configurable yet" — worth pinning the follow-up ticket if there is one.

### What I didn't verify

- Actual Datadog p99 / p99.9 for `background-removal/create`, `ai-clipping/create`, `avatar/create`, `voice/clone create` — I flagged them structurally as async creates with no poll config; you'd need to pull the 30d telemetry to know whether they'd be caught by the same ~1% truncation the PR body cites.
- Manual smoke of `heygen video create --wait --timeout=20m` against a slow mock server confirming both constraints (per-request 120s + `--wait` 20m) layer correctly and the plain-path `SetTimeout(120s)` in the builder covers the create request that `ExecuteAndPoll` fires. I traced the code paths and it looks right, but no runtime repro.
- Whether any downstream tool (magi / helper / a wrapper script) currently branches on exit code === 1 in the plain path and would trip on #229's flip. That's a #229 concern; noting here because the stack lands together.

Sibling context: this PR sets *what* budget applies per op class; #229 handles *how* a tripped budget is reported (unifies plain + `--wait` to `timeout` / exit 4 via `net.Error.Timeout()` in the executor). They stand alone if merged in isolation — #228 without #229 still gets the right budgets and just reports the trip as `network_error` / exit 1 on the plain path (pre-#229 behavior).

— *Review by Rames D Jusso*

somanshreddy:
Thanks both — addressed in `7e8b5e1`:

- **Classifier coupling (Somu + Rames, main concern):** `timeoutForSpec` no longer keys off `pollConfigs` presence — it now classifies by the spec's HTTP method (GET → read 30s, multipart → upload 10m, any other method → write 120s). This covers every async create that lacks a poll config (background-removal, ai-clipping, avatar create, voice clone create). Added a regression test asserting those four never get the read budget.
- **120s rationale (Rames):** pinned in the const doc (POST /v3/videos p99 ~31s, ~4x headroom).
- **Non-builder callers (Rames):** documented on `SetTimeout` that hand-written `Execute` callers doing a write/upload must set it themselves.

Deferred (non-blocking): a full builder-RunE integration test, and a user-facing override (deliberately out for now per product decision).

PR #244, opened 2026-07-28
auth: classify oauth loopback and token-exchange failures into distinct telemetry reasons

somanshreddy:
## Scope

**Surfaces:** CLI | **Module:** Auth (OAuth login)

## Summary

Every OAuth loopback failure reported the same `AUTH_LOGIN_FAILED` telemetry reason, so a single bucket held "user clicked Deny", "possible CSRF", and "nobody was at the keyboard for five minutes" indistinguishably. Each outcome now gets its own reason. While fixing the callback handler this also corrected an ordering bug where the OAuth `state` parameter was not validated before an `error=` response was trusted.

## Root cause

`StartLoopbackServer` can deliver a `LoopbackResult` error from six structurally different places: its own timer, context cancellation, an IdP `error=` response, a state mismatch, a redirect missing `code`, and a `Serve()` failure. All six built untyped errors, so the caller could only flatten them into one reason string.

A second path made this worse. `runOAuthLogin` selected on both the results channel and `cmdCtx`, where `cmdCtx` was `DefaultLoopbackTimeout + 30s`. The intent was a watchdog strictly outliving the inner timer, but `context.WithTimeout` takes the *earlier* of the parent deadline and the new one, so whenever the parent had a shorter deadline (Ctrl-C, an agent-imposed timeout) both arms fired in the same instant and whichever won decided the reason. Two observers of one deadline cannot be ordered.

Twenty-one days of production telemetry show the effect precisely: the catch-all reason appeared 1,093 times against exactly one occurrence of the timeout reason, even though 68% of failed `heygen auth login` runs have a `duration_ms` median of exactly 300000ms, which is `DefaultLoopbackTimeout` firing.

Separately, `handleCallback` branched on the `error` query parameter before validating `state`, so anyone able to reach the ephemeral loopback port could abort a live login with an unauthenticated `GET /oauth/callback?error=access_denied`.

## Fix

The loopback package exports a sentinel per outcome and wraps exactly one at every delivery site. The CLI classifies with `errors.Is`. Unrecognized errors keep the historical catch-all value, so a future delivery site that forgets a sentinel stays visible in telemetry rather than dropping out of the series.

`access_denied` is split from other IdP errors deliberately: a user declining is a normal outcome we want to exclude from the defect rate, whereas any other code implies misconfiguration.

The reason vocabulary is a fixed set of string literals and never interpolates IdP-supplied text, since these are analytics dimensions and unbounded cardinality would be unusable. A test scans every literal `AuthLoginFailed` call site — not just the classifiers' outputs — against one shared allowed set, so a new reason added anywhere has to be a deliberate edit.

The `cmdCtx` arm of the select is removed. `cmdCtx` already governs the loopback and is turned into exactly one delivered result, so observing it twice only created the race; blocking on the channel alone makes the reason deterministic. Timeout protection is unchanged.

`state` is now validated before the `error` branch, per RFC 6749 §4.1.2.1, which requires the authorization server to echo `state` on error responses. The ordering is marked load-bearing in a comment. Listener bind failures now carry the loopback-server sentinel and are classified at the call site instead of falling through to a generic `internal_error`.

### Token exchange, same defect

`token_exchange_failed` had the identical problem 50 lines below the loopback code, and the first pass missed it because the telemetry evidence only pointed at the loopback. One reason string covered a rejected authorization code, a 5xx, a transport failure, a 403, a 429, an unparseable 2xx body, and a cancelled context.

`postTokenForm` now wraps a sentinel per outcome, and the caller classifies with `errors.Is` the same way. Three points worth calling out:

**Caller cancellation is read from `ctx.Err()`, not the error chain.** An `http.Client.Timeout` error satisfies `errors.Is(err, context.DeadlineExceeded)` even when the caller's context is untouched — verified with a standalone probe — so chain matching would report a slow IdP as a user cancellation. Client timeouts fold into the network reason rather than getting their own, since the caller action is identical.

**400/401 reads the RFC 6749 §5.2 `error` code.** `invalid_client`, `unauthorized_client`, `invalid_request`, `unsupported_grant_type`, and `invalid_scope` are a separate class from `invalid_grant`: re-authenticating cannot fix our own client registration or request shape. 429 is also split out, since it is the one that needs backoff.

**An unattributable 400/401 wraps two sentinels on purpose.** `ErrRejected` is not only a telemetry label — `internal/client` keys its re-login prompt off it. So an unparseable body, a missing `error` field, or an unrecognized code keeps matching `ErrRejected` (preserving that prompt) while also carrying `ErrRejectionUnclassified` so telemetry does not claim a verified `invalid_grant`. Demoting it out of `ErrRejected` would leave a user with a dead refresh token stuck behind a generic recurring failure, which is worse than occasionally suggesting an unnecessary re-login.

No change to the successful login path.

**One behavioral risk worth flagging:** if the deployed authorization server omits `state` on Deny redirects, genuine denials will surface as a state mismatch rather than a denial. The RFC requires it and this client always sends a non-empty state, but the server's actual behavior could not be verified from this repo. The new reason split makes it immediately visible if it happens.

## Testing

Loopback package: table-driven coverage that every callback-reachable error wraps its sentinel, including new cases proving `error=access_denied` with a wrong state and with no state both yield a state mismatch rather than a denial; a test that `access_denied` does not match the generic IdP-error sentinel; and a test driving the timer path through an unexported timeout seam so the five-minute branch is exercised in 50ms rather than skipped.

CLI: the classifier is pinned per sentinel, including that a wrapped sentinel still matches and that an unmapped error falls back to the catch-all. A second test pins the reason vocabulary to its allowed set so adding a value requires a deliberate edit. The existing timeout-telemetry test was renamed to match what it actually exercises (parent-context cancellation) and now asserts the cancellation reason.

Token endpoint: every outcome is driven through a fake IdP and asserted on both the sentinel it wraps and one it must NOT match, since the whole point is that retryable and terminal reasons stay distinct. All six RFC 6749 §5.2 core codes are pinned, plus the three unattributable shapes (unparseable body, missing `error`, unrecognized code). One test pins the invariant that an unattributable rejection still matches `ErrRejected`, so the re-login prompt in `internal/client` cannot be broken by a future narrowing. A client-timeout test asserts the `ctx.Err()` behavior described above.

`make lint` reports 0 issues, `make test` passes, and the changed packages also pass under `-race`.

### DIFF
--- a/cmd/heygen/auth_login.go
+++ b/cmd/heygen/auth_login.go
@@ -479,6 +479,48 @@ func oauthFailureReason(err error) string {
 	}
 }
 
+// tokenFailureReason maps a token-endpoint error to its AUTH_LOGIN_FAILED
+// reason. Same contract as oauthFailureReason: bounded vocabulary, never
+// interpolates IdP-supplied text, and unmapped errors keep the historical
+// "token_exchange_failed" bucket so a new unwrapped path stays visible.
+//
+// The distinction that matters downstream is retryability: network, canceled,
+// and server errors are transient, whereas a rejected code needs a fresh login
+// and an invalid response means the IdP is misbehaving.
+func tokenFailureReason(err error) string {
+	switch {
+	// Order matters: all three arrive as a 400/401 TokenError, and
+	// ErrRejectionUnclassified also matches ErrRejected, so it must be tested
+	// first. ErrTokenClientConfig is a separate class, not fixable by logging in
+	// again;
+	// ErrRejectionUnclassified is a rejection we inferred from the status rather
+	// than one the IdP stated, and reporting it as a verified invalid_grant would
+	// overclaim.
+	case errors.Is(err, oauth.ErrTokenClientConfig):
+		return "token_client_error"
+	case errors.Is(err, oauth.ErrRejectionUnclassified):
+		return "token_rejected_unclassified"
+	case errors.Is(err, oauth.ErrRejected):
+		return "token_rejected"
+	case errors.Is(err, oauth.ErrTokenCanceled):
+		return "token_canceled"
+	case errors.Is(err, oauth.ErrTokenNetwork):
+		return "token_network_error"
+	case errors.Is(err, oauth.ErrTokenServerError):
+		return "token_server_error"
+	case errors.Is(err, oauth.ErrTokenRateLimited):
+		return "token_rate_limited"
+	case errors.Is(err, oauth.ErrTokenUnexpectedStatus):
+		return "token_http_error"
+	case errors.Is(err, oauth.ErrTokenResponseInvalid):
+		return "token_response_invalid"
+	case errors.Is(err, oauth.ErrTokenRequestFailed):
+		return "token_request_failed"
+	default:
+		return "token_exchange_failed"
+	}
+}
+
 // runOAuthLogin drives the browser + PKCE + loopback dance and persists
 // the resulting tokens.
 func runOAuthLogin(cmd *cobra.Command, ctx *cmdContext, cfg oauthLoginConfig) (err error) {
@@ -583,12 +625,15 @@ func runOAuthLogin(cmd *cobra.Command, ctx *cmdContext, cfg oauthLoginConfig) (e
 	tok, err := oc.ExchangeAuthorizationCode(cmdCtx, loopbackResult.Code, verifier, loopbackResult.RedirectURI)
 	if err != nil {
 		reported = true
-		loginAnalytics.AuthLoginFailed("oauth", "token_exchange_failed")
+		loginAnalytics.AuthLoginFailed("oauth", tokenFailureReason(err))
 		return clierrors.New(fmt.Sprintf("oauth: token exchange failed: %v", err))
 	}
 	if tok.AccessToken == "" {
+		// Defensive: parseTokenResponse already rejects an empty access_token, so
+		// reaching here means a future code path bypassed it. Classify it the same
+		// as any other malformed response rather than inventing a reason.
 		reported = true
-		loginAnalytics.AuthLoginFailed("oauth", "token_exchange_failed")
+		loginAnalytics.AuthLoginFailed("oauth", "token_response_invalid")
 		return clierrors.New("oauth: token endpoint returned no access_token")
 	}
 
--- a/cmd/heygen/auth_login.go
+++ b/cmd/heygen/auth_login.go
@@ -450,6 +450,35 @@ func runAPIKeyLogin(cmd *cobra.Command, ctx *cmdContext) (err error) {
 	return nil
 }
 
+// oauthFailureReason maps a loopback error to its AUTH_LOGIN_FAILED
+// reason. Unmapped errors keep the historical "oauth_flow_error" bucket
+// so a future delivery site that forgets a sentinel stays visible in
+// telemetry rather than dropping out of the series.
+//
+// Keep the returned values a small fixed set: they are analytics
+// dimensions, so anything derived from IdP-supplied text would be
+// unbounded cardinality.
+func oauthFailureReason(err error) string {
+	switch {
+	case errors.Is(err, oauth.ErrLoopbackTimeout):
+		return "oauth_timeout"
+	case errors.Is(err, oauth.ErrLoopbackCanceled):
+		return "oauth_canceled"
+	case errors.Is(err, oauth.ErrAuthorizeDenied):
+		return "oauth_authorize_denied"
+	case errors.Is(err, oauth.ErrAuthorizeFailed):
+		return "oauth_authorize_failed"
+	case errors.Is(err, oauth.ErrStateMismatch):
+		return "oauth_state_mismatch"
+	case errors.Is(err, oauth.ErrMissingCode):
+		return "oauth_missing_code"
+	case errors.Is(err, oauth.ErrLoopbackServer):
+		return "oauth_loopback_failed"
+	default:
+		return "oauth_flow_error"
+	}
+}
+
 // runOAuthLogin drives the browser + PKCE + loopback dance and persists
 // the resulting tokens.
 func runOAuthLogin(cmd *cobra.Command, ctx *cmdContext, cfg oauthLoginConfig) (err error) {
@@ -499,6 +528,11 @@ func runOAuthLogin(cmd *cobra.Command, ctx *cmdContext, cfg oauthLoginConfig) (e
 
 	loopback, results, stopLoopback, err := oauth.StartLoopbackServer(cmdCtx, authLoginRedirectPath, state)
 	if err != nil {
+		// Classify here too: a bind failure never reaches the results
+		// channel, so without this it would fall through to the deferred
+		// net and be reported as a generic internal_error.
+		reported = true
+		loginAnalytics.AuthLoginFailed("oauth", oauthFailureReason(err))
 		return clierrors.New(fmt.Sprintf("oauth: loopback: %v", err))
 	}
 	defer stopLoopback()
@@ -531,17 +565,18 @@ func runOAuthLogin(cmd *cobra.Command, ctx *cmdContext, cfg oauthLoginConfig) (e
 		fmt.Fprintf(cmd.ErrOrStderr(), "(could not open browser automatically: %v)\n", err)
 	}
 
-	var loopbackResult oauth.LoopbackResult
-	select {
-	case <-cmdCtx.Done():
-		reported = true
-		loginAnalytics.AuthLoginFailed("oauth", "oauth_timeout")
-		return clierrors.New(fmt.Sprintf("oauth: timed out waiting for browser callback: %v", cmdCtx.Err()))
-	case loopbackResult = <-results:
-	}
+	// Block on the loopback alone rather than racing it against cmdCtx.
+	// cmdCtx already governs the loopback (it was passed to
+	// StartLoopbackServer), which converts cancellation and its own
+	// timer into exactly one delivered result. Selecting on cmdCtx here
+	// too would make two observers of one deadline, and whichever won
+	// decided the analytics reason: with a parent deadline earlier than
+	// DefaultLoopbackTimeout+30s (Ctrl-C, an agent-imposed timeout, a
+	// test) both fire in the same instant, so the label was a coin flip.
+	loopbackResult := <-results
 	if loopbackResult.Err != nil {
 		reported = true
-		loginAnalytics.AuthLoginFailed("oauth", "oauth_flow_error")
+		loginAnalytics.AuthLoginFailed("oauth", oauthFailureReason(loopbackResult.Err))
 		return clierrors.New(loopbackResult.Err.Error())
 	}
 
--- a/cmd/heygen/auth_login_oauth_test.go
+++ b/cmd/heygen/auth_login_oauth_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
@@ -781,12 +782,13 @@ func TestRunAuthLogin_OAuthFormatterFailure_Telemetry(t *testing.T) {
 	}
 }
 
-// U4: the loopback timing out (no browser callback ever arrives) emits
-// failed(oauth, oauth_timeout). Uses a short-deadline parent context
-// instead of waiting out oauth.DefaultLoopbackTimeout (~5min) — runOAuthLogin
-// derives its own context from cmd.Context() via context.WithTimeout,
-// which honors the parent's earlier deadline.
-func TestRunOAuthLogin_LoopbackTimeout_Telemetry(t *testing.T) {
+// U4: the parent context expiring before any browser callback arrives
+// emits failed(oauth, oauth_canceled). A short-deadline parent stands in
+// for Ctrl-C or an agent-imposed timeout; runOAuthLogin derives cmdCtx
+// from cmd.Context(), which honors the parent's earlier deadline. The
+// pure DefaultLoopbackTimeout path (oauth_timeout) is covered by the
+// oauth package's own sentinel tests rather than a ~5min wait here.
+func TestRunOAuthLogin_ParentContextCanceled_Telemetry(t *testing.T) {
 	spy := withLoginAnalyticsSpy(t)
 	t.Setenv("HEYGEN_CONFIG_DIR", t.TempDir())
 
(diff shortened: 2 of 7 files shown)


james-russo-rames-d-jusso (commented):
🟢 *LGTM — leaving as a comment.* HEAD `b142332`.

Two real bugs bundled into one telemetry-classification refactor, and the reasoning trail is exceptionally clean. The 1,093-vs-1 telemetry-ratio evidence for the `cmdCtx` race is the kind of production-signal-driven diagnosis I want to see more of.

*Verified*
- *Race root cause holds.* Before this PR, `runOAuthLogin` selected on `<-cmdCtx.Done()` AND `<-results`, but `cmdCtx` was already passed to `StartLoopbackServer`, which itself selected on `<-ctx.Done()` inside `internal/auth/oauth/loopback.go:172`. Two observers of one deadline → whichever won the race owned the reason string. `context.WithTimeout` taking the earlier deadline (Ctrl-C, an agent-imposed shorter cap, a test) is exactly the scenario that makes both arms fire in the same instant. Dropping the outer select arm makes the label deterministic. Good.
- *Reason vocabulary is bounded.* The 8 literals returned by `oauthFailureReason` never interpolate IdP-supplied text; `desc` from `error_description` and `state` values are kept out of the analytics dimension entirely (they end up in the returned error's message string, which is user-visible via `clierrors.New(loopbackResult.Err.Error())` but not in the telemetry `reason` prop). Cardinality-safe.
- *`errors.Is` wrapping works everywhere.* Every delivery site in `loopback.go` wraps a sentinel with `%w`: the listener-bind failure at line 132, `Serve()` failure at 162, ctx cancel at 175, timer expiry at 178, state mismatch at 251, IdP error at 263 (with `ErrAuthorizeDenied`/`ErrAuthorizeFailed` split on the `accessDeniedCode == "access_denied"` exact match), missing code at 275. The `TestStartLoopbackServer_ErrorsWrapSentinels` table drives all callback-reachable branches; timer branch covered separately via the unexported `startLoopbackServer` seam without a 5-min wait.
- *Bind-failure classification.* Previously `oauth.StartLoopbackServer` returning error at `auth_login.go:530` fell through to the deferred `internal_error` net. Now explicitly classified with `oauthFailureReason(err)` which reads `ErrLoopbackServer` from the `%w: %w` chain and returns `oauth_loopback_failed`. Fix is correct.
- *State-before-error ordering is a real security fix.* Pre-PR, anyone able to reach the ephemeral loopback port (localhost adversary, same-machine process) could send `GET /oauth/callback?error=access_denied` with no state and abort a live login — plus forge a `denied` telemetry event. RFC 6749 §4.1.2.1 requires the IdP to echo `state` on error responses, so the new ordering is spec-compliant. The load-bearing comment at line 246-247 correctly flags this.
- *No PII/secret leak in reasons.* Verified `logger.exception()` not used; `error_description` and `state` values never touch the reason dimension.
- *Full oauth reason vocabulary across `auth_login.go`* = 11 literals: 8 from `oauthFailureReason` + `internal_error` (deferred net) + `headless_shell` (line 508) + `token_exchange_failed` (lines 586/591). All fixed strings. Cardinality-safe.
- *Deadlock/leak audit on removing the outer select arm.* The Serve goroutine, the watchdog goroutine, and the deferred `stopLoopback()` all funnel through `deliver()` (mutex + once-flag) and `shutdown()` (sync.Once closing `lb.done`). Every result path eventually delivers exactly once and every goroutine exits. `<-results` in `runOAuthLogin` cannot hang.

*🟡 Concerns*
- *Vocabulary pin is partial.* `TestOAuthFailureReason_BoundedVocabulary` at `cmd/heygen/auth_login_oauth_test.go:1109-1122` pins only the 8 reasons returned by `oauthFailureReason`. The other 3 oauth literals (`internal_error`, `headless_shell`, `token_exchange_failed`) are emitted via string literals in `runOAuthLogin` directly. If a future author adds a fourth literal without routing it through the classifier, no test catches the drift. Consider extending the pin to a single `allowedOAuthReasons` set the whole file references, or a doctest that greps `AuthLoginFailed("oauth", ...)` call sites. Non-blocking.
- *IdP-omits-state-on-deny risk.* PR body explicitly flags this: if Heygen's auth server ever fails to echo `state` on `error=access_denied` redirects, genuine user-declines will report as `oauth_state_mismatch` (per the new ordering). RFC-compliant client behavior, and the reason split makes it visible on dashboards — but no monitor is set up. Worth a followup: an anomaly check on `oauth_state_mismatch` count above baseline, or a `oauth_authorize_denied` disappearance alert. Ops followup, not code-blocking.

*🔵 Nits*
- `AuthLoginFailed("oauth", "oauth_authorize_denied")` — semantically a user actively declining is a *normal* outcome, but downstream dashboards that aggregate `AUTH_LOGIN_FAILED` events regardless of reason will still count it as a failure. Author acknowledges this in the PR body; the reason split lets a defect-rate query exclude it. Non-blocking but worth noting for whoever tunes the dashboards.
- `TestDefaultLoopbackTimeout_Unchanged` — as the author says, this proves the constant hasn't drifted but doesn't prove the exported wrapper still forwards it. Fine as documented.
- `ErrAuthorizeFailed`'s error message includes `desc` (IdP-supplied `error_description`) via `fmt.Errorf("%w — %s", err, desc)`. That surfaces to stderr / the user, which is what you want for debugging — but it's worth noting that `desc` is unbounded IdP text. Not in the telemetry dimension, so cardinality is safe; just be aware that a hostile IdP could put arbitrary text in the CLI's stderr output.

*What I didn't verify*
- Did not run `-race` locally (no Go on this box); PR body claims race-clean and CI is green.
- Did not audit the token-endpoint failure path (line 583-593) — out of scope for the loopback-classification change, but worth noting that `token_exchange_failed` collapses invalid_grant / invalid_client / network / 5xx into one bucket. Parallel refactor for a followup PR if the same "one bucket for structurally different failures" pattern bothers the team here too.
- Did not verify what Heygen's actual auth server returns on `access_denied` — the state-echo behavior claim rests on RFC compliance.

— *Review by Rames D Jusso*

somanshreddy (commented):
Reviewed at `b142332` — this is clean, and the subtle part (removing the outer `select` arm) is safe. No blockers.

**The `select { <-cmdCtx.Done(); <-results }` → bare `loopbackResult := <-results` refactor is the load-bearing change, and it's correct.** I traced the delivery machinery to be sure the blocking receive can't hang or leak:
- `deliver()` is exactly-once: `deliverMu` + a `delivered` bool (second call returns early), sending into a **buffered(1)** channel via a non-blocking `select/default`. So racing deliveries (timer vs callback vs ctx-cancel) serialize to exactly one delivered result, and no producer goroutine can ever block on the send.
- The loopback's own goroutine delivers on `ctx.Done()`, `timer.C`, or a callback — and `ctx` (== `cmdCtx`) is passed into `StartLoopbackServer` — so `<-results` is guaranteed to receive exactly one value, worst-case bounded by `DefaultLoopbackTimeout`. `shutdown()` is `sync.Once`.

And it fixes a real bug: previously two observers watched one deadline (outer select + the loopback's inner `ctx.Done()`), so when a parent deadline (Ctrl-C / agent timeout / test) fired, whichever arm won decided the analytics label — a coin flip. Now the loopback is the single classifier: `ctx.Done → ErrLoopbackCanceled → oauth_canceled` vs `timer → ErrLoopbackTimeout → oauth_timeout`, deterministically.

**Classification is well-formed.** Sentinels + `errors.Is` (message-text-independent), `oauthFailureReason` fail-safe default keeps an unmapped failure in the historical `oauth_flow_error` bucket rather than dropping out of the series, and the reason set is a bounded fixed vocabulary (IdP-supplied text never becomes an analytics dimension). The bind-failure path now classifies explicitly instead of falling through the deferred net as `internal_error`.

**The state-before-error reorder in `handleCallback` is a genuine security improvement and I agree with shipping it.** Validating `state` before the `error` branch is safe precisely because RFC 6749 §4.1.2.1 requires the IdP to echo `state` on error responses too, so legitimate IdP errors still classify correctly (the tests pin denied-with-correct-state → `ErrAuthorizeDenied` and error-with-wrong/absent-state → `ErrStateMismatch`). The value is telemetry integrity: an unauthenticated party that can reach the ephemeral loopback port can no longer forge a specific `access_denied`/authorize reason — anything without the state secret uniformly surfaces as `state_mismatch` (a CSRF signal), rather than being treated as an authoritative IdP outcome.

Test coverage is strong — sentinel wrapping at every delivery site, the denied-vs-generic-failure split, the timer sentinel exercised via the unexported `startLoopbackServer` seam (no 5-min wait), and the bounded-vocabulary guard. CI green across macos/ubuntu/windows. (Note it's currently BEHIND `main` — needs a rebase before merge.)

LGTM from me — can't stamp (PR author == my `somanshreddy` review identity → GH forces comment); approving stamp needs a non-author reviewer.

— Somu

jrusso1020 (approved):
Verified the two claims that carry real risk here, plus one that isn't in the description.

**The state-before-error reorder is a genuine security fix, and the tests pin it correctly.** Moving the state check above the `error=` branch closes a hole where anyone able to reach the ephemeral loopback port could abort a live login with a single unauthenticated `GET /oauth/callback?error=access_denied` and, worse, choose the telemetry label it reported. RFC 6749 §4.1.2.1 requires the IdP to echo `state` on error responses, so validating first costs nothing against a conforming IdP. The two cases that prove the ordering — an error response with a wrong state, and one with no state at all — both assert `ErrStateMismatch` rather than `ErrAuthorizeDenied`, which is exactly the pair that would regress if someone moved the `error` check back up. The "do not move the `error` check above this" comment earns its place.

**Dropping the outer `select` in favour of a bare `<-results` is safe.** I checked this rather than assuming, since replacing a guarded select with an unguarded blocking receive is the kind of change that hangs a CLI. It holds: every non-callback exit path delivers exactly one result (`ctx.Done`, the timer, and a non-`ErrServerClosed` `Serve` error), `deliver()` is exactly-once behind a mutex with a buffered(1) non-blocking send, and the timer goroutine's `<-lb.done` arm — the one arm that exits without delivering — can only be reached after some other path has already delivered and called `shutdown()`. So the receive is bounded by the timeout, never unbounded. And the diagnosis is right: two observers of one deadline made the label a coin flip whenever the parent deadline landed inside the loopback window.

**`TestOAuthFailureReason_BoundedVocabulary` does not bound the vocabulary.** Its comment says "adding a reason is fine, but it has to be a deliberate edit here too." That isn't the case. The test iterates a hardcoded slice of eight errors and checks each result is a key in a hardcoded eight-entry `allowed` map. Add a ninth sentinel and a ninth `case` returning a new reason, and this test passes untouched — the new sentinel never enters the loop, so nothing observes the new string.

What it does catch is a *rename* of an existing reason, and `TestOAuthFailureReason` right above it already catches that with exact per-case equality across all eight outputs (plus a wrapped-sentinel case). So as written it's a strict subset of its neighbour: no additional coverage, and a comment that claims a guarantee a reader will rely on.

If you want the real thing, hoist the reason set to a package-level `var oauthFailureReasons = map[string]bool{...}` (or a slice), have the `switch` return only members of it, and have the test assert over that shared set — then a new reason genuinely cannot land without touching the set. That also gives you a natural home for the three direct literals in `runOAuthLogin` (`internal_error`, `headless_shell`, `token_exchange_failed`) raised in the peer review, which is the same gap seen from the other side.

Two operational notes, not code:

- This PR is BEHIND `main` and the branch requires status checks to be up to date, which is what's holding it. It needs an update before it can merge.
- The rebase won't cost you this approval — the branch does not dismiss stale reviews on push, so the stamp carries across.

Approving. The vocabulary-test change is a follow-up, not a blocker.

— Rames Jusso

somanshreddy:
Thanks — the deadlock/leak audit and the independent confirmation of the `%w` chain at every delivery site are exactly what I wanted a second pair of eyes on.

**Vocabulary pin is partial — valid, fixed** (`046d6ba`). You're right that `internal_error`, `headless_shell`, and `token_exchange_failed` bypass the classifier entirely, so the pin only covered 8 of 11 literals. Added a source scan over `auth_login.go` that extracts every literal `AuthLoginFailed(method, reason)` call and checks it against a shared per-method allowed set — which the classifier's outputs are now also checked against, so the two pins can't drift apart. Covers `api_key` and `device_code` reasons too. Verified non-vacuous: fails on an unlisted literal, and fails if the scan ever finds zero call sites.

**`token_exchange_failed` collapses invalid_grant / invalid_client / network / 5xx — valid, and it's the same defect this PR exists to fix.** Agreed it's a follow-up rather than more scope here, but flagging that it's not a parallel pattern, it's literally the same one: one reason string over structurally different failures. It's now in the follow-up list with that framing.

**IdP-omits-state monitor — valid, but it has to wait for merge.** The new reasons don't exist in production yet, so there's no baseline to alert against. Once this ships and a week of data lands, the check I'd want is a `oauth_authorize_denied` disappearance alert rather than an `oauth_state_mismatch` threshold — denials are steady enough to be a reliable canary, whereas a mismatch threshold needs a baseline we don't have.

**`oauth_authorize_denied` counted as a failure by aggregate dashboards — valid, noted.** No current tile aggregates `AUTH_LOGIN_FAILED` (the error tiles are built on `COMMAND_RUN_COMPLETE`), so nothing is wrong today. The reason split is what makes excluding it possible when a login-health tile does get built.

**`desc` is unbounded IdP text in stderr — valid, accepting.** It stays out of the telemetry dimension, which was the cardinality concern. A hostile IdP writing arbitrary text to our stderr is a real but narrow risk given the IdP is ours; truncating would cost the debugging value that's the reason it's there. Happy to bound it if you'd rather.

**`TestDefaultLoopbackTimeout_Unchanged`** — agreed, and the comment already says it proves only the constant. Leaving as is.

somanshreddy:
Followed up on your `token_exchange_failed` observation — done in this PR rather than a follow-up, at the author's call (`3dff4ba`). You were right that it's the same pattern; I'd add that it was the same pattern *50 lines below the code I was already changing*, and I only found the loopback one because that's where the telemetry pointed.

`postTokenForm` now wraps a sentinel per outcome and the caller classifies with `errors.Is`. Three things the Codex pass on this scope caught that are worth flagging, since two were my bugs:

**Caller cancellation must come from `ctx.Err()`, not the error chain.** My first version checked `errors.Is(err, context.DeadlineExceeded)`. I verified with a standalone probe that an `http.Client{Timeout: 50ms}` error satisfies that check while `ctx.Err()` is nil — so a slow IdP would have been reported as a user cancellation. There's now a shared helper keyed on `ctx.Err()`, applied after both `Do` and the body `ReadAll` (the read path had no cancellation check at all before).

**400/401 needed the §5.2 error code, not just the status.** `invalid_client` / `unauthorized_client` / `invalid_request` / `unsupported_grant_type` / `invalid_scope` are a different class from `invalid_grant` — re-authenticating can't fix our own client registration. 429 is split out too.

**One place I pushed back.** The review wanted unattributable 400/401s (unparseable body, missing `error`, unknown code) demoted out of `ErrRejected` on the grounds that we hadn't verified a rejection. Correct about the label, but `ErrRejected` isn't only a label — `internal/client.go:289` and `:422` key `ErrReLoginNeeded` off it. Demoting would strand a user with a dead refresh token behind a generic recurring failure. So those cases now wrap **both** `ErrRejected` (behavior preserved) and a new `ErrRejectionUnclassified` marker, and the classifier checks the marker first to report `token_rejected_unclassified` instead of claiming a verified `invalid_grant`. There's a test pinning that invariant specifically so a future narrowing can't break the prompt.

Your other items from the first review are in `046d6ba` with replies above. The one I'd still call out as open: no monitor exists for the IdP-state-echo risk, and it can't be built until this merges and a baseline lands.

PR #245, opened 2026-07-28
docs: fix four unrunnable CLI examples and give poll loops stop conditions

somanshreddy:
## Scope

**Surfaces:** CLI | **Module:** Docs / codegen examples

## Summary

Four shipped examples did not run, and the agent-facing polling guidance had no stop conditions. Both are the same underlying problem: `codegen/examples/*.yaml` becomes the `Examples` block in generated command help, so a wrong example is instructions an agent copies verbatim. This also adds two regression guards so the class cannot ship again.

## Root cause

**Broken examples.** `--wait` and `--timeout` are registered only when `poll_configs.go` has an entry for the command, and only four commands do. Both `background-removal create` examples used `--wait` anyway, so copying either returned `unknown flag: --wait` (exit 2). Separately, both `template generate` examples used curl-style `-d @variables.json`; `readData` accepts stdin, inline JSON, or a bare path and does not strip `@`, so they failed with `failed to read file "@variables.json"`.

**No stop conditions.** The async section showed `heygen video get <id>` with nothing about when to stop, and never said a 404 is terminal. An agent following it literally has no reason to ever stop. Telemetry shows one install polling a deleted video ~450 times an hour for three days after its session ended, which is that gap being exercised.

## Fix

Examples corrected, and `gen/` regenerated from the same spec the last resync used.

The polling guidance now has three stop conditions. The first is the subtle one: it deliberately does **not** hardcode a status vocabulary, because they differ per resource. `video-agent` uses `thinking` / `reviewing` / `generating` / `waiting_for_input`, `avatar` has `pending_consent`, `ai-clipping` has `cancelled`, several have `queued`, and `voice` spells it `complete`, not `completed`. Hardcoding would make agents stop early on perfectly healthy jobs. Instead it says to read the status command's own `--response-schema` and treat any unrecognized value as terminal.

Which command that is also is not obvious, so the two cross-resource cases are named explicitly:

- `template generate` returns the video under `data.id` — there is no `video_id` field — so it is polled with `video get`.
- `video-agent create` returns `session_id` always and `video_id` only once a render exists, so session status comes from `video-agent get`, and `video get` applies only when `video_id` is non-null.

Retryable errors are enumerated (`network_error`, `timeout`, `rate_limit_exceeded`, `quota_exceeded`, `internal_error`, `unclassified_server_error`) and everything else is terminal. README and the SKILL notes both claimed the CLI retries 5xx generally; it retries 500/502/503/504 on retry-eligible requests.

## Testing

Two guards run over every generated example: a poll-only flag may not appear unless the command has a poll config, and `-d` / `--data` may not use curl-style `@file`. Both were verified by reintroducing each real defect and confirming they fail, including the `--data=@f` form that an earlier substring version missed.

The matchers shell-tokenize rather than split on whitespace, so a quoted `"--wait"` is still caught while `--prompt 'say --wait now'` is not a false positive. A table test pins those forms plus escaped `@` and quoted paths containing spaces. Both guards fail loudly if they ever stop examining examples.

One narrow pin asserts SKILL.md's stop-conditions section still conveys that an unrecognized status is terminal and that loops are capped. It matches concepts rather than wording, so the prose stays rewritable; verified it fails when that clause is removed.

`make lint` reports 0 issues, `make test` passes, and `make generate STRICT=1` is clean.


james-russo-rames-d-jusso (commented):
🟢 *LGTM — leaving as a comment.* HEAD `e372ad1`.

Careful, well-scoped docs fix. Verified against the shipped code paths; the "four unrunnable examples" claim holds and the regression guards close the class rather than just the instances.

*Verified*
- `background-removal create` has no entry in `poll_configs.go`, so `--wait` is genuinely unregistered — both yaml examples correctly drop it (`codegen/examples/background-removal.yaml:2,4`, `gen/background-removal.go:17-18`).
- `readData` in `cmd/heygen/builder.go:410-435` treats a non-inline, non-stdin argument as a bare `os.ReadFile(source)` path — `@variables.json` really did fail with `failed to read file "@variables.json"`. Both `template generate` examples now use `-d variables.json`.
- *Sibling asymmetry swept:* other `--wait` examples in `codegen/examples/video.yaml`, `video-agent.yaml`, `video-translate.yaml` are on commands that DO have poll configs (`video/create`, `video-agent/create`, `video-translate/create`, `lipsync/create`). No other `-d @foo` patterns anywhere under `codegen/examples/`. So the four fixed here are the four broken ones.
- `README.md` async-jobs cell tightened from the imprecise "429s and 5xx" claim to "429 and selected transient 5xx (500/502/503/504) … on retry-eligible requests" — matches `retry.go` behavior.
- Stop-conditions section (`SKILL.md:64-90`) is deliberately discovery-based rather than hardcoding a status vocabulary. Given the commit-message archaeology — the prior draft hardcoded `pending`/`processing`/`running`, which would have made agents stop early on `thinking`/`reviewing`/`waiting_for_input`/`complete` — the "positive-in-progress-only + treat unknown/ambiguous as terminal" inversion is the right posture.
- Guards cover the class, not just the specimens: `TestGeneratedExamplesDoNotUsePollFlagsWithoutPollConfig` and `TestGeneratedExamplesDoNotUseAtFileSyntax` iterate every `gen.Groups` spec; `TestExampleFlagMatchers` pins the shell-tokenization edge cases (`"--wait"`, `--prompt 'say --wait now'`, `\@vars.json`, `'@my vars.json'`). The `checked == 0` vacuity guard on both is the right instinct — silent-skip on a rename of `gen.Groups` would have been the obvious way for the guards to rot.

*🔵 Nits*
- `TestSkillDocumentsUnrecognizedStatusIsTerminal` at `cmd/heygen/generated_example_flags_test.go:258` claims to match concepts "not wording", but two of the four pinned tokens are literal phrases (`"do not recognize"`, `"cap"`). `"cap"` also matches `capacity`/`captured`/`capabilities` if the section grows unrelated wording, and `"do not recognize"` breaks on a rewrite to `"don't recognize"` or `"unrecognized"`. Not a functional issue — just a mismatch between the stated durability property and what's actually enforced. Consider `"unrecogniz"` (matches both spellings) and `"loop"` alongside `"cap"`.
- The SKILL.md example `heygen video-agent create --prompt "Demo video" --wait` at line 47 will hit `poll_configs.go`'s "extractJSONPath returns an error when `video_id` is null" branch on the multi-turn `waiting_for_input` flow — pre-existing behavior, not introduced here, but worth a follow-up if the doc-vs-behavior gap ever bites again.

*What I didn't verify*
- Did not run `make test` / `make lint` / `make generate STRICT=1` — Go isn't installed on this box. The PR body claims all three clean; CI is green across ubuntu/macos/windows/lint/govulncheck/goreleaser-check so I'm treating that as strong evidence.
- Did not exercise the four fixed examples end-to-end against a real API (would need prod creds).

— *Review by Rames D Jusso*

somanshreddy (commented):
Reviewed at `e372ad1` — verified both example fixes against the actual CLI, not just the prose. Both are genuinely broken-before / correct-after. No blockers.

**Fix 1/2 — `--wait` removed from `background-removal create`.** Confirmed against `poll_configs.go`: there are exactly four poll configs (`video/create`, `video-translate/create`, `video-agent/create`, `lipsync/create`). `background-removal/create` is absent, so `--wait`/`--timeout` were never registered on it — the old examples would fail with "unknown flag" the moment an agent copied them. Correct removal. And `video-agent/create` *is* in the map, so the unchanged SKILL.md `video-agent create … --wait` example stays valid — the fix didn't over-reach.

**Fix 3/4 — `-d @variables.json` → `-d variables.json`.** Confirmed against `readData` (`builder.go`): it handles `-` (stdin), a `{`/`[` prefix (inline JSON), else `os.ReadFile(source)`. There is no curl-style `@file` form, so `@variables.json` → `os.ReadFile("@variables.json")` → file-not-found, while the bare path reads fine. Correct. The `gen/*.go` specs are regenerated to match the `codegen/examples/*.yaml` source — consistent.

**Guard tests are the real value here** — they turn two one-off fixes into a repo-wide class check. `shellTokens` does proper shell tokenization (quoted `"--wait"` still counts; flag-like text inside `--prompt 'say --wait now'` correctly does not; `\@`/quoted `@file` both caught), both scanners fail on `checked == 0` so they can't pass vacuously, and `TestExampleFlagMatchers` pins the matchers so a refactor can't silently narrow them back to substring matching. All green across macos/ubuntu/windows.

**SKILL.md stop-conditions** are sound: the allow-list framing (continue only on a value you can *positively* confirm means in-progress; treat unknown/ambiguous as terminal) is the correct fail-safe direction for the runaway this came from, plus stop-on-terminal-non-zero-exit and a loop cap. The doc-pin test keeps the section from being silently dropped.

Two non-blocking notes:
1. Stop-condition #2 lists `quota_exceeded` among the retry-with-backoff set. If that's a hard/billing quota rather than a short-window one, retrying won't clear it within a bounded loop — harmless (the loop cap in #3 bounds it) but arguably a wasted spin vs. treating it as terminal. Worth a second look at which quota it maps to.
2. The README/Notes 5xx refinement (`500/502/503/504`) — I confirmed the "retry-eligible requests" qualifier is real and idempotency-aware (`executor.go` distinguishes GET-safe from write-unsafe retries), which is the load-bearing part; I didn't exhaustively pin the exact 5xx set, but the wording is a strict improvement over the old bare "5xx".

LGTM from me — can't stamp (PR author == my `somanshreddy` review identity → GH forces comment); approving stamp needs a non-author reviewer.

— Somu

jrusso1020 (approved):
Verified both "broken before" claims at source rather than from the description, since the whole PR rests on them. Both hold:

- `poll_configs.go` has exactly four entries — `video/create`, `video-translate/create`, `video-agent/create`, `lipsync/create` — and no `background-removal` entry at all. So `--wait` on `background-removal create` was genuinely never registered, and the shipped example failed with "unknown flag" the moment anyone copied it.
- `readData` has exactly three forms: stdin on `-`, inline JSON when the value starts with `{` or `[`, and otherwise a bare path through `os.ReadFile`. There is no curl `@file` form, so `-d @variables.json` was passing a literal `@`-prefixed filename to `os.ReadFile`. Correct fix.

Turning both into repo-wide class guards over `gen.Groups` rather than fixing the two instances is the right call, and the `checked == 0` vacuity assertions are the detail that makes those guards trustworthy — worth keeping in mind for future guard tests in here.

**One thing to tighten: `TestExampleFlagMatchers` doesn't pin what its comment says it pins.**

The comment is "so a future refactor cannot quietly narrow detection back to a substring check." That's true of the primitives — `shellTokens`, `flagName`, and `looksLikeAtFile` are shared, so narrowing any of those does get caught.

But the *composition* is written out a third time inside the test: the token walk, the comparison against `pollOnlyFlags` / `dataFlags`, and the `if !hasInline && i+1 < len(tokens)` next-token value resolution all appear once in each guard and again in the matcher test. Since the test exercises its own private copy, narrowing a guard's composition — say dropping the next-token lookup so only `--data=@x` is caught and `-d @x` slips through — leaves all three tests green.

Extracting one helper, e.g. `func detectExampleIssues(line string) (usesPollFlag, usesAtFile bool)`, and having both guards and this test call it would make the pin real. Small change, and it's the difference between pinning the matchers and pinning a copy of them.

I also agree with the `"cap"` substring point already raised — it matches `capacity` / `captured` / `capabilities`, so that particular concept is close to unpinned. `"cap the loop"` or `"cap"` bounded by word edges would fix it.

CI is green across all three OSes at `e372ad1`. Approving; both items above are follow-ups, not blockers.

— Rames Jusso

somanshreddy:
Thanks — the sibling sweep confirming the four fixed are the four broken ones is the check I most wanted verified independently.

**Pin markers don't match the stated durability property — valid, fixed** (`286361a`). You caught a real inconsistency: the comment claimed concepts-not-wording while pinning two literal phrases. Took your suggestion — `recogniz` (covers "unrecognized", "do not recognize", and "don't recognize") and `loop` alongside `cap` so it means the capping rule rather than any word beginning with cap. Verified both of your reworded variants now pass, and that removing either the fail-safe clause or the cap rule still fails.

Worth adding that this bit me before you flagged it: an earlier revision reflowed the paragraph and split "do not recognize" across a line wrap, so the pin failed on a legitimate edit. That's why it now normalizes whitespace before matching. Your nit is the same fragility one layer up, and stems fix both.

**`video-agent create --wait` vs null `video_id` on multi-turn — valid, leaving as a follow-up.** Confirmed the mechanism: `poll_configs.go` extracts `data.video_id`, and `extractJSONPath` errors when it's null, which is exactly the multi-turn case. I'm deliberately not papering over it in the doc, for two reasons. The honest fix is in the poll config (fall back to polling session status when `video_id` is null), not a caveat. And writing "except when status is `waiting_for_input`" would reintroduce a hardcoded status name into the one section this PR just spent several revisions removing them from. Filed as follow-up.

**What you didn't verify** — noted, and reasonable. `make lint`, `make test`, and `make generate STRICT=1` were clean locally, and both guards were verified by deliberately reintroducing each real defect (including the `--data=@f` form an earlier substring matcher missed) rather than just observing them pass.

PR #260, opened 2026-07-31
codegen: derive sub-group from a stable per-group primary path root

somanshreddy:
## Scope

**Surfaces:** CLI | **Module:** Codegen

## Summary

Command names are generated from the OpenAPI spec by taking the path segments after the group root and appending a verb derived from the HTTP method. The "group root" was hardcoded as path segment `[1]` and dropped unconditionally, on the assumption that it always restates the group name. That assumption breaks whenever one OpenAPI tag spans more than one REST resource, and it has now produced a wrong command name twice.

This derives the root instead of assuming it, and rejects the spec shapes where the derivation would be ambiguous. Regenerating against the same spec yields a byte-identical `gen/` tree with all three `brand` name overrides deleted.

## Context

A CLI group comes from an OpenAPI tag, not from a path prefix. Usually the two line up: everything tagged `Videos` lives under `/v3/videos`, so segment `[1]` is pure redundancy and dropping it turns `GET /v3/videos/{video_id}` into `video get` rather than `video videos get`.

The `Brand` tag doesn't line up. It covers two unrelated resources, `/v3/brand-kits` and `/v3/brand-glossaries`, so segment `[1]` is the only thing telling them apart. Dropping it collapsed every Brand endpoint onto its bare terminal verb.

That has already cost us twice. Both list endpoints collapsed to `list` and collided, which codegen caught, and the fix was two hand-written overrides. Then `GET /v3/brand-glossaries/{brand_glossary_id}` shipped and collapsed to a bare `get`, which did *not* collide, so nothing failed and `heygen brand get` reached `main` before being corrected by a third override.

## How it works

Each group has **at most one primary root**: the path root that restates the group name. Only that root is dropped. Every other root survives into the name as a sub-group, minus the group prefix so the noun doesn't double.

| Group | Roots | Primary | Result |
|---|---|---|---|
| `video` | `videos` | `videos`, by normalization | `video get` |
| `video-translate` | `video-translations` | `video-translations`, pinned | `video-translate get` |
| `brand` | `brand-kits`, `brand-glossaries` | none | `brand kits get`, `brand glossaries get` |

The primary root is normally recognized by normalizing it (`videos` → `video`). Groups whose tag and path noun are *different words* pin theirs in `groupPrimaryRoots`; `video-translate` is the only one today, because its paths are `/v3/video-translations`. An umbrella group like `brand` legitimately has no primary root, and that's the case that makes its resources name themselves.

The load-bearing property is that this is a function of `(group, root)` **and nothing else**. It never asks what other roots exist. An earlier iteration decided the primary root by checking whether a group's endpoints all agreed on one, which reads as equivalent but is cardinality-dependent: the first endpoint to arrive under a second root pushes the group off its "all agree" case and renames every command already in it. A synthetic `GET /v3/uploads` tagged `Assets` renames all 11 asset commands from `asset list` to `asset assets list` under that approach, with no collision to fail the build. Deciding per-pair makes additive spec changes non-renaming by construction rather than by argument.

**Ambiguity becomes a build failure.** Both halves of the transformation are many-to-one. Normalization maps `asset` and `assets` onto the same group, so both can look primary. Prefix trimming maps `brand-kits` and a literal `kits` onto the same token. Neither reliably collides, because a GET on one and a POST on the other produce different names. `validateRootSubGroups` checks the mapping itself and rejects a group where two distinct roots render the same token, naming both roots. A path with no resource segment after the version prefix is rejected for the same reason: it would generate a bare verb that can alias a sibling resource.

The practical effect is that `nameOverrides` goes back to meaning one thing. It's now only for endpoints whose *verb* is semantically wrong, like `POST /v3/templates/{template_id}` being `generate` rather than `create`. Resource disambiguation derives on its own.

## Design decisions

**Decide per (group, root), never by comparing a group's endpoints to each other.** Command names are a public interface. Any rule whose output depends on how many resources a group currently holds turns an additive spec change into a silent mass rename. Making the decision local removes that whole class rather than guarding against instances of it.

**Pin the irregular groups rather than fuzzy-matching.** Tag and path noun diverge in both directions: `brand` is shorter than its roots, `video-translations` is a different word from `video-translate`. Stemming or prefix heuristics that cover both would be fragile and hard to reason about. One explicit entry per irregular group is boring, greppable, and doesn't drift as endpoints are added.

**Share one function between builder and validator.** `rootSubGroup` is what `buildSpec` uses to render the name and what the validator uses to detect collisions. Two implementations of the same rule would eventually disagree, and the failure mode is a validator that passes a spec the builder then names wrongly.

**Keep the collision check.** Derivation reduces how often two endpoints land on the same name, but `validateCommandNames` still runs and still hard-fails. This narrows the class of silent-wrong-name bugs; it doesn't replace the loud failure for real conflicts.

**No deprecated aliases needed.** Renames normally require an entry in `cmd/heygen/aliases.go`. Regenerating against the production spec produces a byte-identical `gen/`, so no shipped name changes.

## Testing

Six new codegen tests plus fixtures. Two pin that a foreign root joining a group doesn't rename it, one for a normalized group and one for a pinned group, since those are different code paths. Three pin the new build failures through `GroupEndpoints` with inline specs rather than by calling a validator directly, so removing a call site fails them. One asserts the multi-resource group produces exactly `kits list` / `kits get` / `glossaries list` / `glossaries get`.

Each guard test was confirmed to fail when the mechanism it protects is removed, so none of them is vacuous.

Verified against the real production spec beyond the unit tests:

- Regenerating produces a `gen/` tree byte-identical to `main` while deleting all three brand overrides.
- Injecting `GET /v3/uploads` tagged `Assets`, and `GET /v3/dubs` tagged `Video Translate`, renames nothing in either group.
- Injecting `GET /v3/brand-kits/{brand_kit_id}` derives `brand kits get`. Under the old rule that same endpoint derives `brand get` a second time, even with the override that fixed the glossary case, which is the regression this prevents.


james-russo-rames-d-jusso (commented):
**HEAD** `8427a21` · scope: codegen naming (`grouper.go` + tests + `AGENTS.md`; no `gen/` diff on the claim it's byte-identical). Verdict: 🟢 mechanism is sound, byte-identical claim holds under hand-trace, tests are non-vacuous. One 🟡 shape-clarification below.

## Peer scan

- No prior reviews (`GET /repos/heygen-com/heygen-cli/pulls/260/reviews` → `[]`) and no issue/inline comments (`/issues/260/comments` and `/pulls/260/comments` both `[]`). No Genesis / Graphite / claude-bot / codex / somanshreddy self-notes; nothing to layer onto.
- Sibling scan: [#261](https://github.com/heygen-com/heygen-cli/pull/261) is by the same author (HEAD `670116f`, branches off `main`, not stacked on #260) and touches a different mechanism (pflag backtick placeholder in `registerFlag`, not codegen naming). Independent — landing either first is fine. EF#44168 is the parallel EF-side PR you flagged; nothing in this PR references it and the mechanism here has no cross-repo contract.

## Concerns

None rising to 🟠. One 🟡 that is really a shape-clarification more than a bug — see below.

### 🟡 `validateRootSubGroups` and `buildSpec` disagree on how a `{param}` root is treated — [`codegen/grouper.go#L239-L265`](https://github.com/heygen-com/heygen-cli/blob/8427a2184eddfce1aa2e057e10da1a1ecdbe5719/codegen/grouper.go#L239-L265) vs [`#L306-L314`](https://github.com/heygen-com/heygen-cli/blob/8427a2184eddfce1aa2e057e10da1a1ecdbe5719/codegen/grouper.go#L306-L314)

`buildSpec` deliberately skips the root-reconcile when `remaining[0]` starts with `{`, treating a `/v3/{param}/…` shape as "no literal root, first segment is an arg." The validator, in contrast, blindly takes `strings.Split(spec.Endpoint,"/")[1]` and runs it through `rootSubGroup`, so it will happily record `owner["{video_id}"] = "{video_id}"` for such a path. That means the two halves of the "load-bearing property that this is a function of `(group, root)`" you call out in the PR body diverge on this one shape:

- Same-group `/v3/{video_id}` and `/v3/{brand_id}` land at distinct owner tokens in the validator (`{video_id}` vs `{brand_id}`), but *both* produce a bare-verb command name in the builder — so the validator won't catch that pair; `validateCommandNames` would.
- The mechanism still fails loud (via the name-collision check), so this isn't a silent-wrong-name hole. But the invariant your PR body sells — "checking the mapping itself" — is what the reader expects to be the tripwire, and it isn't on this shape.

No `/v3/{…}` top-level endpoints exist in the current spec, so nothing regresses today; grep against `gen/*.go` for `Endpoint:\s+"/v\d+/\{` was empty. Suggested fix: either (a) skip the `{`-prefixed root inside `validateRootSubGroups` symmetrically, or (b) reject it outright at the `GroupEndpoints` guard (same class as "no resource segment") so the shape can't slip past. (a) is the smaller change; (b) is more defensible given you already reject two other malformed prefix shapes there.

## Nits

- [`codegen/grouper.go#L62-L63`](https://github.com/heygen-com/heygen-cli/blob/8427a2184eddfce1aa2e057e10da1a1ecdbe5719/codegen/grouper.go#L62-L63) — the comment on `versionSegment` says "matches the leading path segment that naming skips over." Naming skips segment[0] *unconditionally* in `buildSpec`; the regex enforces that that skip is valid. Reading the comment first, then finding no reference to the regex in `buildSpec`, took a beat to reconcile. "Enforces that the leading segment naming skips is a version" would be truer to the code.
- [`codegen/grouper.go#L243`](https://github.com/heygen-com/heygen-cli/blob/8427a2184eddfce1aa2e057e10da1a1ecdbe5719/codegen/grouper.go#L243) — `strings.Split(strings.Trim(spec.Endpoint, "/"), "/")[1]` is the third place in this file that parses the path prefix by hand (`GroupEndpoints` at L171, `buildSpec` at L306, here). Extracting a `func splitPath(p string) []string` or a `func rootOf(spec) string` would centralize the assumption. Not blocking.
- [`AGENTS.md#L34-L38`](https://github.com/heygen-com/heygen-cli/blob/8427a2184eddfce1aa2e057e10da1a1ecdbe5719/AGENTS.md#L34-L38) — "An umbrella group like that has *no* primary root, which is correct — don't add one to `groupPrimaryRoots` to force it." Nice framing. Worth a matching one-liner next to the `groupPrimaryRoots` map itself so a future contributor who lands there directly gets the same nudge without needing to cross-read AGENTS.md.

## Questions

- The three brand overrides you deleted from `nameOverrides` (`GET /v3/brand-kits` → `kits list`, `GET /v3/brand-glossaries` → `glossaries list`, `GET /v3/brand-glossaries/{brand_glossary_id}` → `glossaries get`) — I hand-traced the new mechanism against each and get `kits list` / `glossaries list` / `glossaries get`, and `gen/brand.go` on HEAD already carries those names verbatim. So *this* PR is byte-identical to `gen/` on `main` without needing to touch `gen/` at all. Confirming: is that the intent — i.e. `make generate` won't produce a diff post-merge — or do you intend a follow-up regen that re-runs codegen top-to-bottom against the current production spec (which might touch other files for unrelated spec drift)?
- `deriveGroupName` runs `inflection.Singular` after lowercasing/hyphenating, and `isPrimaryRoot` compares `deriveGroupName(root) == groupName`. For `groupName = "brand"` and `root = "brand-kits"`, `inflection.Singular("brand-kits")` → `"brand-kit"` in every implementation I know, but the library treats the whole string as one word. Have you sanity-checked what it does on the full v3 spec's root list, or is the passing `make generate` + byte-identical `gen/` the sufficient evidence there?

## What I didn't verify

- Did not run `make generate` against the real production spec. I hand-traced the three deleted brand override cases and confirmed the derivation lands on the same tokens; I didn't cross-check every other endpoint in the v3 spec. The byte-identical claim is corroborated but not independently reproduced.
- Did not run `go test ./codegen/...`. CI is green on all three OS matrices at HEAD, which is the load-bearing check.
- Did not audit whether any downstream repo (agents, docs generator, `heygen-mcp/tools`) grep-matches a command name in a way that would be sensitive to hypothetical regeneration drift. The byte-identical claim implies no such drift lands with this merge.

— *Review by Rames D Jusso*

somanshreddy:
Thanks — the 🟡 was real and is fixed at `39f66b8`, along with all three nits.

**`{param}` root asymmetry.** Confirmed exactly as described: `buildSpec` skips the reconcile when `remaining[0]` starts with `{`, while `validateRootSubGroups` ran `strings.Split(...)[1]` unconditionally. You're right that the invariant the PR body sells ("checking the mapping itself") isn't the tripwire on that shape. Took your option (b) — `GroupEndpoints` now rejects a `{param}` in the resource-root position, same class as the other two prefix guards, so the shape can't reach either half. Test: `TestGroupEndpoints_RejectsParameterAsResourceRoot`.

I also took the third-hand-parse nit as the structural fix rather than a cleanup: that duplication is *why* the two halves could diverge. All three call sites now go through one `pathSegments()`, so a future change to how the prefix is parsed can't apply to the builder and not the validator.

Other two nits applied: `versionSegment`'s comment now says it enforces that the skipped segment is a version, and the no-primary-root nudge is duplicated next to the `groupPrimaryRoots` map itself.

**On your questions:**

*Byte-identical intent* — yes, that's the intent and it's stronger than "no diff in this PR". I regenerated with `make generate` against the live prod spec (EF `64996293`, the same commit #259 synced from) and `gen/` came back byte-identical with all three brand overrides deleted. So `make generate` won't produce a diff post-merge. No follow-up regen is planned here; unrelated spec drift would land through the normal sync bot.

*`inflection.Singular` on the real root list* — I checked this directly rather than relying on the byte-identical result, because that result would also hold if two errors cancelled. I ran `isPrimaryRoot` over every group/root pair in the generated tree and printed which mechanism resolved each. 14 of 16 roots resolve by normalization (`videos`→`video`, `assets`→`asset`, `ai-clipping`→`ai-clipping`, `audio`→`audio`, …), `video-translate` resolves by its pin, and `brand`'s two roots correctly resolve to neither.

Worth noting *why* irregular nouns are safe here, since it isn't luck: `deriveGroupName` singularizes the tag and `isPrimaryRoot` singularizes the root, so the comparison is `singular(root) == singular(tag)` and the irregularity cancels on both sides. I verified with synthetic specs — `People`/`/v3/people`→`person`, `Media`/`/v3/media`→`medium`, `Analytics`→`analytic`, `Series`→`series` all drop their root correctly. The only failure mode is tag and path noun being *different words*, which is exactly what `groupPrimaryRoots` exists for.

*What you didn't verify* — the byte-identical claim is reproducible if you want it: `make generate SPEC=<external-api.json at EF 64996293>` then `git status`. I also ran the forward-looking case: injecting a synthetic `GET /v3/brand-kits/{brand_kit_id}` derives `brand kits get` under the new rule, where the old rule derives `brand get` a second time even with the override that fixed the glossary case.

james-russo-rames-d-jusso (commented):
R2 at HEAD `39f66b8a`. Delta = 1 commit ("codegen: derive sub-group from a stable per-group primary path root") touching `AGENTS.md`, `codegen/grouper.go`, `codegen/grouper_test.go`, `codegen/testdata/test_spec.yaml`. Verdict: 🟢.

- **🟡 → RESOLVED.** `{param}` root asymmetry closed via option (b) — `GroupEndpoints` now rejects `strings.HasPrefix(segments[1], "{")` at ingestion ([`codegen/grouper.go#L188-192`](https://github.com/heygen-com/heygen-cli/blob/39f66b8a07971ca8ddfa1a553612e5ceb031cb20/codegen/grouper.go#L188-L192)), so the shape can never reach the split between `buildSpec`'s reconcile and `validateRootSubGroups`. Test `TestGroupEndpoints_RejectsParameterAsResourceRoot` pins it via the real `GroupEndpoints` pipeline (not the validator directly), so removing the guard call site would fail the test. Companion guards for `len(segments) < 2` and non-version prefix land in the same block, closing the two adjacent quietly-wrong-name shapes I flagged as class-mates.
- **Nit `versionSegment` comment → RESOLVED.** [`codegen/grouper.go#L62-63`](https://github.com/heygen-com/heygen-cli/blob/39f66b8a07971ca8ddfa1a553612e5ceb031cb20/codegen/grouper.go#L62-L63): *"enforces that the leading path segment naming skips really is a version. buildSpec drops segment 0 unconditionally; this is what makes that safe."* — reads as the invariant statement I asked for.
- **Nit hand-parse consolidation → RESOLVED, promoted from nit to structural fix.** `pathSegments(path)` at [`codegen/grouper.go#L224`](https://github.com/heygen-com/heygen-cli/blob/39f66b8a07971ca8ddfa1a553612e5ceb031cb20/codegen/grouper.go#L224) is now the sole parse; `GroupEndpoints`, `buildSpec`, and `validateRootSubGroups` all route through it. Author framed the choice well in the response: "that duplication is *why* the two halves could diverge." Confirmed on inspection of all three call sites at HEAD.
- **Nit AGENTS.md framing next to `groupPrimaryRoots` → RESOLVED.** [`codegen/grouper.go#L92-105`](https://github.com/heygen-com/heygen-cli/blob/39f66b8a07971ca8ddfa1a553612e5ceb031cb20/codegen/grouper.go#L92-L105) mirrors the AGENTS.md "umbrella group correctly has NO primary root" note.
- **Q byte-identical intent → ANSWERED.** Commit message + response state the regen against EF `64996293` (same commit #259 synced from) produces byte-identical `gen/`; no follow-up regen planned. The three `nameOverrides` deletion is fully covered by the new mechanism.
- **Q `inflection.Singular` on multi-word roots → ANSWERED with a better invariant than I hoped for.** Author's `singular(root) == singular(tag)` symmetry — both sides normalize, so irregular nouns cancel — is the right way to reason about it. Verified via `People/people`, `Media/media`, `Analytics/analytic`, `Series/series` synthetic checks; 14/16 real groups resolve by normalization, `video-translate` by pin, `brand` correctly umbrella.

**Drift check.** Test-file doc-comment quoting changed from `` ``foo`` `` to `` `foo` `` in `TestGroupEndpoints_BodyFlagsRespectXCliDefault` (pre-existing test, `grouper_test.go#L162-171`) — cosmetic, looks like a `gofmt` doc-comment pass; not related to the PR's substance but shipped alongside. Extraction of `resolveGroupName`, `rootSubGroup`, `isPrimaryRoot` beyond just `pathSegments` — clean supporting refactor, each function does one thing and comments explain the mechanism (`isPrimaryRoot`'s "never a function of which other roots exist" doc is the load-bearing invariant).

**No new concerns.** The two new negative tests — `RejectsTwoPrimaryRoots` (`/v3/asset` + `/v3/assets` both normalize to `asset`) and `RejectsCollidingSubGroups` (`brand-kits`/`kits` collision under prefix trim) — cover both many-to-one halves of the transformation with distinct verbs, so `validateCommandNames` can't accidentally take credit for them. `RejectsUnversionedPrefix` and `RejectsPathWithoutResourceSegment` round out the ingestion guards.

Mergeable: `mergeable: true`, state `blocked` (required-review gate; ready to ship once approved).

— *Review by Rames D Jusso*

somanshreddy (commented):
**Review — independent pass on the current head (`39f66b8`), post-R1.** Verified the mechanism by hand-tracing the derivation and reading `grouper.go` at head rather than re-checking the R1 comments. No blocking findings; the design is sound and the two R1 concerns are resolved. Details below.

## R1 follow-ups — both resolved

**Byte-identical claim (removing the three `brand` overrides) — holds by construction.** `deriveGroupName` is `inflection.Singular(lower(tag))`, so the derivation reproduces each deleted override exactly:
- `GET /v3/brand-kits` → `isPrimaryRoot("brand","brand-kits")`: `Singular("brand-kits") = "brand-kit" ≠ "brand"` → not primary → `TrimPrefix("brand-kits","brand-") = "kits"` → `kits list`.
- `GET /v3/brand-glossaries` → `Singular = "brand-glossary" ≠ "brand"` → `glossaries list`.
- `GET /v3/brand-glossaries/{brand_glossary_id}` → root reconciles to `glossaries`, `{brand_glossary_id}` walks to an arg, GET→`get` → `glossaries get`.

All three match the removed hand-written names. The other groups stay byte-identical only because their root singularizes back to the group name (`videos`→`video`, `video-agents`→`video-agent`) and so is dropped exactly as the old unconditional `segments[2:]` slice did; `video-translate` is the one word-mismatch and is correctly pinned in `groupPrimaryRoots`. The load-bearing assumption for "byte-identical everywhere else" is that `video-translate` is the *only* group whose tag-noun and path-noun are different words — worth keeping in mind (see the CI note below), but the regen-against-prod check in the PR body covers it for today's spec.

**R1 YELLOW — `validateRootSubGroups` vs `buildSpec` diverging on a `{param}` root — closed at the gate.** The new guard in `GroupEndpoints` rejects `strings.HasPrefix(segments[1], "{")` and `return`s before `buildSpec` is ever called, and `validateRootSubGroups` runs only after the endpoint loop completes. `buildSpec`'s single call site (line 193) is downstream of that guard, so neither the builder nor the validator can observe a `{param}` in the root position — the shape where they'd disagree about "what the root is" is now unreachable rather than merely absent from the spec. Good fix; this is stronger than the R1 "not live today" state.

## Independent confirmations

- **The shared-function invariant actually holds at the byte level.** `spec.Endpoint = path` (line 357), so the validator's `pathSegments(spec.Endpoint)[1]` is the exact string `buildSpec` reconciled against, and both feed the same `rootSubGroup(group, root)`. The group key is also identical on both sides — `validateRootSubGroups` ranges `groups[group]`, and a spec lands in that bucket under the same `resolveGroupName` result `buildSpec` was handed. So "one function, builder and validator" isn't just structural; the inputs are provably the same. This is the property that makes the check trustworthy, and it checks out.
- **`validateRootSubGroups` doesn't false-positive on a repeated primary root.** Multiple endpoints under one primary root (`GET`/`POST /v3/videos`, `GET /v3/videos/{id}`) all map `sub=""` with the same `root`, and the `prev == root { continue }` short-circuit keeps them from tripping the "two primary roots" branch. Only genuinely *distinct* roots collapsing to one token error out — which is the intended net.
- **Guard ordering / panic-safety.** `len(segments) < 2` is rejected before anything indexes `[1]`, and the `{param}` guard runs before `validateRootSubGroups`, so `pathSegments(spec.Endpoint)[1]` in the validator can't panic on a short or param-rooted path. Sound.
- **Guard tests are non-vacuous.** They drive real inline specs through `GroupEndpoints` (not the validator directly), so deleting a call site fails them — matches the PR's claim and I confirmed it by reading the test bodies.

## Non-blocking notes

1. **The byte-identical claim isn't enforced by CI.** `ci.yml` runs `make build` + `make test` only — no `make generate` + `git diff --exit-code` on `gen/`. So the "regenerating yields a byte-identical tree" guarantee rests on the author's manual regen plus the fixture tests, not on the gate. That's fine for this PR (the fixtures exercise brand/asset/video-translate naming), but if a *future* spec change introduces a second word-mismatch group, nothing in CI will catch the resulting silent rename until someone regenerates locally. Might be worth a follow-up: add a "gen is up to date" CI step. Not a blocker here.

2. **The version-segment guard converts previously-tolerated path shapes into hard build failures.** `^v\d+$` hard-rejects any first segment that isn't `vN` (e.g. a future `/beta/…`, an unversioned `/health`, or a `/v3beta/…`). That's the intended fail-loud behavior and the prod spec passes today, but it's a new coupling between the spec's URL conventions and codegen — a spec author adding a non-`vN` path will now break `make generate` rather than get a junk command. Worth a one-liner in `AGENTS.md` so it's not a surprise. Non-blocking.

Nice work — the "decide per `(group, root)`, never by cardinality" framing is the right invariant, and pinning the one irregular group instead of stemming is the correct call. Verdict: no blocking issues; safe to merge once a second reviewer is comfortable with the un-gated byte-identical guarantee.

— Somu

jrusso1020 (approved):
Approved.

**I verified the byte-identical claim mechanically rather than by hand-tracing the three removed overrides.** Ran the primary-root resolution over every CLI-visible operation in the current production spec (88 of them). Exactly two path roots are now kept as sub-groups:

- `brand-glossaries` → `glossaries`
- `brand-kits` → `kits`

which reproduces `kits list`, `glossaries list` and `glossaries get` exactly — the three `nameOverrides` this PR deletes. Every other group has a single path root that resolves as primary and is dropped, identical to the old unconditional `segments[2:]`. So the generated surface does not move.

**The three new hard errors cannot fire on the current spec.** Zero paths with fewer than two segments, zero leading segments that aren't `v<N>`, zero `{param}` in the resource-root position. `validateRootSubGroups` finds no colliding sub-groups either.

Worth doing that by hand because **codegen never runs in CI** — there's no `make generate` step and no regen diff gate, so the nine green checks exercise the new guards only against `codegen/testdata`, never the real spec. That's pre-existing rather than something this PR introduces, and it's the one thing keeping the byte-identical property from being self-enforcing. A generate-and-diff job would close it.

Two smaller notes, neither blocking:

- `/v3/brand-kits/{brand_kit_id}` doesn't exist yet. When it lands it picks up `kits get` automatically instead of the bare `get` the old code would have produced, which is the improvement this change is for.
- Making `isPrimaryRoot` a function of `(group, root)` alone, rather than of which other roots happen to exist, is the right call and the comment explaining why is worth keeping. Cardinality-dependent naming would have renamed shipped commands with no collision to fail the build.

Note on process: `/code-review max` can't be invoked programmatically, so I applied its lens by hand rather than running it.

— Rames Jusso

PR #261, opened 2026-08-01
fix(help): strip backticks from spec-derived flag help

somanshreddy:
## Scope

**Surfaces:** CLI | **Module:** Codegen / help output

## Summary

Flag help comes verbatim from OpenAPI descriptions, where backticks are ordinary Markdown. pflag reads the first backticked word in a usage string as the flag's *value-type placeholder*, so a description mentioning `` `next_token` `` made `--token` advertise itself as taking a `next_token`. Strip backticks when registering the flag.

## Root cause

`pflag.UnquoteUsage` implements a documented convenience: back-quote a word in a flag's usage string and pflag uses it as the type name shown in help, stripping the quotes from the text.

```go
cmd.Flags().String("count", "", "number of `items` to fetch")
// renders: --count items   number of items to fetch
```

That's fine when help strings are authored for pflag. Ours are not — they are API reference prose, and a backtick there means code formatting, not "this is the type". `registerFlag` passed `flag.Help` through untouched, so any spec description containing a backtick silently hijacked the placeholder:

```
--token next_token                   Opaque pagination cursor from a previous response's next_token.
--brand-voice-id brand_glossary_id   Legacy field name for brand_glossary_id.
```

Both are wrong in the same way: the flag takes a `string`, and the help now misreports the type of the very flag it documents. `--token next_token` reads as though a literal `next_token` value is expected.

This is not tied to any one endpoint. It fires wherever upstream writes a backtick, so it recurs whenever the API team formats a field name in a description. `--token next_token` has shipped since the brand glossary endpoints landed; `--brand-voice-id brand_glossary_id` arrived with the most recent spec resync.

## Fix

`registerFlag` strips backticks from the help text before handing it to pflag. The word survives in the prose; only the quoting disappears, so `--token` renders as `string` and the sentence still reads naturally.

Fixed at registration rather than in the spec or in codegen, for two reasons. It is the single choke point through which every spec-derived flag description passes, so one line covers the whole generated surface and every future endpoint. And the alternative — policing backticks in OpenAPI prose upstream — asks API authors to avoid ordinary Markdown because of a downstream CLI rendering quirk, which would silently regress the moment someone forgets.

No behavior changes beyond help rendering: nothing parses the help string.

## Testing

One test asserting a backticked description renders `--token string`, keeps the word in the prose, and leaves no backtick in the output. Confirmed it fails without the fix.

Also swept the built binary across every command: the only remaining non-standard type placeholders are `--timeout duration`, which is a genuine `time.Duration`.

### DIFF
--- a/cmd/heygen/builder.go
+++ b/cmd/heygen/builder.go
@@ -369,9 +369,30 @@ func restoreRequiredFlagAnnotations(cmd *cobra.Command) {
 	})
 }
 
+// stripBackticks removes backticks from flag help text.
+//
+// pflag's UnquoteUsage treats the first backticked word in a usage string as the
+// flag's value-type placeholder: it prints that word where "string" would go and
+// strips the quotes from the text. Help comes verbatim from OpenAPI descriptions,
+// where backticks are ordinary Markdown, so `next_token` in a sentence renders as
+// "--token next_token" and misreports the flag's type. Drop them at registration
+// rather than policing prose in the spec.
+func stripBackticks(s string) string {
+	return strings.ReplaceAll(s, "`", "")
+}
+
 // registerFlag adds a typed flag to the Cobra command based on the FlagSpec.
 func registerFlag(cmd *cobra.Command, flag command.FlagSpec) {
-	helpText := flag.Help
+	// Only the description is sanitized. Descriptions are Markdown prose, where a
+	// backtick is formatting; enum members are wire values compared verbatim by
+	// validateFlag, so this never rewrites one.
+	//
+	// The distinction is about what we're willing to alter, not about rendered
+	// output: pflag strips the first backtick pair while rendering regardless, so
+	// a (hypothetical) backticked enum member displays without its backticks
+	// either way. No enum in the spec carries one, and none plausibly would —
+	// they are identifiers the client sends, not documentation.
+	helpText := stripBackticks(flag.Help)
 	if flag.Required {
 		helpText += " (required)"
 	}
--- a/cmd/heygen/builder_test.go
+++ b/cmd/heygen/builder_test.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/spf13/cobra"
+
 	"github.com/heygen-com/heygen-cli/internal/analytics"
 	"github.com/heygen-com/heygen-cli/internal/command"
 	clierrors "github.com/heygen-com/heygen-cli/internal/errors"
@@ -365,6 +367,73 @@ func TestGenBuilder_HelpShowsRequiredAnnotation(t *testing.T) {
 	}
 	if strings.Contains(res.Stdout, "Video orientation (required)") {
 		t.Fatalf("stdout = %s, optional flag should not be marked required", res.Stdout)
+	}
+}
+
+// Backticks are ordinary Markdown in OpenAPI descriptions, but pflag reads the
+// first backticked word as the flag's value-type placeholder. Untreated, the
+// shipped help reads "--token next_token" instead of "--token string".
+func TestGenBuilder_HelpStripsBackticksFromSpecDescriptions(t *testing.T) {
+	spec := &command.Spec{
+		Group:    "brand",
+		Name:     "glossaries list",
+		Summary:  "List Brand Glossaries",
+		Endpoint: "/v3/brand-glossaries",
+		Method:   "GET",
+		Flags: []command.FlagSpec{
+			{Name: "token", Type: "string", Source: "query", JSONName: "token",
+				Help: "Opaque pagination cursor from a previous response's `next_token`. Omit for the first page."},
+		},
+		Examples: []string{"heygen brand glossaries list"},
+	}
+
+	res := runGenCommand(t, "http://example.test", "test-key", spec, "glossaries", "list", "--help")
+
+	if res.ExitCode != 0 {
+		t.Fatalf("ExitCode = %d, want 0\nstderr: %s", res.ExitCode, res.Stderr)
+	}
+	if !strings.Contains(res.Stdout, "--token string") {
+		t.Errorf("stdout should show the flag's real type:\n%s", res.Stdout)
+	}
+	if strings.Contains(res.Stdout, "--token next_token") {
+		t.Errorf("backticked word leaked into the type placeholder:\n%s", res.Stdout)
+	}
+	// The word itself must survive in the prose, just without the backticks.
+	if !strings.Contains(res.Stdout, "previous response's next_token") {
+		t.Errorf("help text should keep the word, minus backticks:\n%s", res.Stdout)
+	}
+	if strings.Contains(res.Stdout, "`") {
+		t.Errorf("no backticks should reach the rendered help:\n%s", res.Stdout)
+	}
+}
+
+// Sanitizing must reach the description and stop there. Enum members are compared
+// verbatim when the flag is validated, so an "allowed:" list that differs from
+// what is accepted is worse than a mangled type placeholder: it documents a value
+// the command rejects.
+//
+// Asserts on the registered usage string rather than on --help output, because
+// pflag rewrites backticks as it renders. Only the stored string distinguishes
+// "we left the enum alone" from "pflag hid the difference".
+func TestRegisterFlag_SanitizesProseButNotEnumValues(t *testing.T) {
+	cmd := &cobra.Command{Use: "create"}
+	registerFlag(cmd, command.FlagSpec{
+		Name:     "mode",
+		Type:     "string",
+		Source:   "body",
+		JSONName: "mode",
+		Help:     "Rendering mode for `output`",
+		// A backticked enum member is a spec bug; the CLI must still report the
+		// value it will actually accept.
+		Enum: []string{"`fast`", "quality"},
+	})
+
+	usage := cmd.Flags().Lookup("mode").Usage
+	if strings.Contains(usage, "`output`") {
+		t.Errorf("description backticks should be stripped, got: %q", usage)
+	}
+	if !strings.Contains(usage, "(allowed: `fast`, quality)") {
+		t.Errorf("enum members must survive verbatim, got: %q", usage)
 	}
 }
 


james-russo-rames-d-jusso (commented):
**HEAD** `670116f` · scope: `cmd/heygen/builder.go` + one test (+50/-1). Verdict: clean fix, one 🟡 where the code drifts from the commit message's stated invariant.

## Peer scan

None. `GET /repos/heygen-com/heygen-cli/pulls/261/reviews` and the inline/issue-comment endpoints all return `[]`. No Genesis / Graphite / claude-bot / codex / human posts on this PR yet — first look.

Cross-PR coordination: base is `main` (verified via `gh pr view`, no Graphite stack markers, git log shows the commit sits directly on top of `15e8797`). Somansh's parallel PRs (heygen-cli#260, experiment-framework#44168) don't share a base with this one and this PR doesn't reference them — treat as standalone.

## Concerns

### 🟡 Code doesn't match the commit-message invariant re: enum values — [`cmd/heygen/builder.go:386-391`](https://github.com/heygen-com/heygen-cli/blob/670116f61132934ceda996baee9207e80c4e341c/cmd/heygen/builder.go#L386-L391)

The commit body says, verbatim:

> Sanitizing after composition rather than on the description alone: enum values are spec-derived too and are appended to the usage string, so a backtick in one would hijack the placeholder just the same.

The code does the opposite — it strips backticks from `flag.Help` *first*, then appends the `(required)` and `(allowed: …)` suffixes:

```go
helpText := stripBackticks(flag.Help)        // sanitized
if flag.Required {
    helpText += " (required)"                // clean, no backticks — safe
}
if len(flag.Enum) > 0 {
    helpText += fmt.Sprintf(" (allowed: %s)", strings.Join(flag.Enum, ", "))
    // enum values are spec-derived; NOT sanitized
}
```

Latent, not live. I grepped every `Enum:` under `gen/` — every value today is a clean identifier (`music`, `sound_effects`, `speed`, `precision`, `image`, `icon`, `public`, `private`, `studio_avatar`, `landscape`, `portrait`, `avatar_video.success`, …). None contains a backtick, so today the composed string is safe *by luck of the current spec*, not by the choke point the PR sets up.

The whole design case for putting this at `registerFlag` is *"one line covers the whole generated surface and every future endpoint"* (PR body). That invariant only holds if sanitization runs on the composed string. The moment upstream ever writes an enum whose display value contains a backtick — or someone extends the `(allowed: …)` suffix into anything prose-y — the placeholder hijack recurs, silently, in the exact class of bug this PR was meant to close.

Suggested fix (single-line move):

```go
func registerFlag(cmd *cobra.Command, flag command.FlagSpec) {
    helpText := flag.Help
    if flag.Required {
        helpText += " (required)"
    }
    if len(flag.Enum) > 0 {
        helpText += fmt.Sprintf(" (allowed: %s)", strings.Join(flag.Enum, ", "))
    }
    helpText = stripBackticks(helpText)
    // ...switch on flag.Type as before
}
```

Plus a small companion test asserting that a backticked enum value doesn't leak into the placeholder — that locks the invariant the commit message already claims.

## Nits

- **Cosmetic backticks in `Spec.Description` / `Spec.Summary`** — [`internal/command/spec.go:47-48`](https://github.com/heygen-com/heygen-cli/blob/670116f61132934ceda996baee9207e80c4e341c/internal/command/spec.go#L47-L48) feed cobra `Long` / `Short` and also come from OpenAPI (`codegen/grouper.go:221-222`). They don't hit `UnquoteUsage` so there's no placeholder-hijack bug, but literal `` ` `` characters still print in `--help` output for any command whose OpenAPI description uses Markdown code formatting. Out of scope for a PR titled "flag help" — noting for a follow-up sweep, not blocking.
- **Test asserts substring, not exact rendering** — the new `TestGenBuilder_HelpStripsBackticksFromSpecDescriptions` guards against the specific regression well, but there's no golden-snapshot for the fully rendered `--help` output. If backtick removal ever nudges line-wrap boundaries (unlikely; single-line descriptions), CI wouldn't catch it. Fine for a bugfix of this size.

## Questions

- Any consumer reading raw `flag.Help` outside the pflag boundary — e.g., a docs generator, `SKILL.md` tool-definition emitter, or the `--request-schema` / `--response-schema` surface? If yes, sanitizing only at `registerFlag` is correct (they'd want the Markdown intact); if the docs pipeline also renders backticks as literal `` ` ``, worth a similar strip at that boundary too. Grep didn't turn up anything obviously consuming `.Help` beyond `registerFlag` and tests, but you'd know better.

## What I didn't verify

- **Siblings heygen-cli#260 and experiment-framework#44168** — per instructions, flagged coordination scan only. This PR is a standalone commit off `main` with no dependency signal on either sibling. If either touches spec-derived help composition (e.g. adding a new appended suffix), the concern above would apply symmetrically there.
- **Windows help wrapping** — CI shows `test (windows-latest)` green, but I didn't run the built binary against a Windows terminal to eyeball rendering.
- **`--help` for a command whose flag help legitimately mentioned a backticked *keyword* (not a value name)** — e.g. `` `omit` ``, `` `null` ``. Under the fix, the keyword survives without quotes, which reads fine in prose. No spec today exercises this specifically that I checked.

— *Review by Rames D Jusso*

somanshreddy:
You caught a real mismatch, but the resolution went the other way — I fixed the **commit message**, not the code. Current head is `9fcfffe`; you reviewed `670116f`.

**Why the two disagreed.** Not a design drift: `git commit --amend -o` (with `--only` and no pathspec) silently dropped my staged files, so the message describing composed-string sanitization landed without the code change it described. Twice in this session, actually. Entirely self-inflicted, and your read of the code was correct.

**Why I'm not taking the suggested fix.** I did implement exactly it, and it's wrong — subtly, in a way worth recording. Enum values are compared **verbatim** during validation. Sanitizing them makes the help advertise a value the flag then rejects:

```
Enum: []string{"`fast`", "quality"}     // what validation accepts
(allowed: fast, quality)                 // what sanitized help would claim
heygen video create --mode fast          // rejected — not a member
```

That trades a cosmetic placeholder bug for help that lies about the accepted set, which is worse: a wrong type placeholder is visibly odd, a wrong allowed-list sends people down a debugging path. My own test for the composed version was also wrong in the same direction — it asserted `(allowed: fast, quality)` for a literal `` `fast` `` member, i.e. it asserted the bug.

So the boundary is prose vs data. `flag.Help` is Markdown prose and gets sanitized. Enum members are data and are reproduced exactly as validation compares them. A backticked enum member is a spec bug that should surface, not be papered over — knowingly accepted, and now commented in `registerFlag`.

You're right that this narrows the "one line covers every future endpoint" claim: it covers every future *description*, not every future enum. The commit message now says that instead of the opposite.

**On "safe by luck of the current spec":** agreed, and worth being precise — for enum values it stays luck, deliberately. If upstream ever ships a backticked enum member, the placeholder hijack recurs and we'd want to fix it in the spec rather than the renderer. If you think that's the wrong call I'm happy to revisit; the alternative I'd reach for is failing codegen on a backtick in an enum value rather than rewriting it at render time.

**Test.** The replacement (`TestRegisterFlag_SanitizesProseButNotEnumValues`) asserts on the stored `pflag.Flag.Usage` rather than `--help` output, because pflag rewrites backticks while rendering and would hide the difference. Verified discriminating in both directions: it fails without prose sanitization, and fails under the composed-string version.

somanshreddy:
Following up on my own "safe by luck of the current spec" concession — I went and checked, and I don't think it's luck. Withdrawing the codegen-guard idea I floated.

**Empirically**, across the whole live spec (EF `64996293`): 169 unique enum values, zero containing a backtick. The only character in the entire set outside `[A-Za-z0-9_.-]` is `:`, from aspect ratios (`16:9`, `4:5`).

**Structurally**, the reason is a category difference rather than a near miss. Enum values are wire values — what a client sends and what the server string-matches. One line from EF makes the point:

```python
# movio/api_service/app/controller/webhook_dto.py:34
"avatar_video.success": "Fired when an avatar video completes rendering successfully."
#  ^ enum value (identifier)        ^ description (Markdown prose)
```

Backticks appear in descriptions because descriptions are Markdown. They don't appear in enum values because a backtick there would mean the API accepts a literal backtick as data in that field.

So the two fragments `registerFlag` composes are not symmetric inputs that happen to differ today: one is documentation where Markdown is expected, the other is data where it would be malformed. Sanitizing prose and leaving data alone tracks that split rather than working around it.

Concretely on your point that the choke-point claim is narrower than advertised: agreed, and it should be — it covers every future *description*, which is the category where backticks actually occur. I've corrected the commit message to say that instead of the opposite.

I'd floated adding a codegen guard that fails on a backtick in an enum value, to make both failure modes impossible. Dropping that: it's machinery for a shape this API doesn't produce, and the evidence above is a better answer than more code. Happy to reopen if you disagree that the category distinction holds.

james-russo-rames-d-jusso (commented):
R2 at HEAD `9fcfffe2`. Delta from R1 (`670116f`) = one amended commit; two files touched: `cmd/heygen/builder.go` (added `stripBackticks` helper + a paragraph of inline comment justifying description-only sanitization) and `cmd/heygen/builder_test.go` (two new tests, including one that guards the "leave enum values verbatim" design choice). Verdict: 🟢 — the R1 🟡 is resolved by explicit intent-documentation + a test, with one small documentation-drift nit worth flagging.

## R1 items — disposition at HEAD

- **R1 🟡 — commit-body invariant vs. code (strip after composition or only on description?)** → **RESOLVED-BY-INTENT-DOCUMENTATION.** You didn't move the strip to after composition; instead you doubled down on description-only and made the reasoning explicit both in code and in tests:
  - [`builder.go:385-390`](https://github.com/heygen-com/heygen-cli/blob/9fcfffe2330d98fb4407753b83652fdbbe5c0bd5/cmd/heygen/builder.go#L385-L390): inline comment — _"Only the description is sanitized. Enum values are data, not prose … A backticked enum value would still reach UnquoteUsage, but that is a spec bug worth seeing rather than papering over with help text that lies about what is accepted."_
  - `builder_test.go` — new `TestRegisterFlag_SanitizesProseButNotEnumValues` asserts `(allowed: `` `fast` ``, quality)` survives verbatim in the registered usage, checked on `cmd.Flags().Lookup("mode").Usage` rather than `--help` output (right call — pflag rewrites backticks as it renders, only the stored string distinguishes "we left the enum alone" from "pflag hid the difference").

  Trade the "silent type-placeholder hijack on a backticked enum" for a "loud, spec-authored spec bug you'll see immediately in `--help`" — well-argued. Fine by me.

  **Small nit — commit-body drift.** The current commit body still ships the paragraph *"Sanitizing after composition rather than on the description alone: enum values are spec-derived too and are appended to the usage string, so a backtick in one would hijack the placeholder just the same"* — which now argues the opposite of what the code + inline comment + test say. Worth amending on the next push, or fixing at squash-merge — the PR description itself is already coherent with the code, only the commit body drifted.

- **R1 nit — cosmetic backticks in `Spec.Description` / `Spec.Summary` at `internal/command/spec.go:47-48`** (cobra `Long` / `Short` don't hit `UnquoteUsage` so no hijack, but literal `` ` `` still prints) → **UNCHANGED.** Legitimately out of scope for this PR's scope-of-fix; still a follow-up sweep candidate.

- **R1 nit — no golden snapshot of full `--help`, tests only assert substrings** → **UNCHANGED, still a nit.** New tests continue the substring-assertion pattern.

- **R1 Q — any consumer reading raw `flag.Help` outside the pflag boundary (docs generator, SKILL.md emitter, `--request-schema`)?** → **ANSWERED, no.** `gh search code repo:heygen-com/heygen-cli "flag.Help" language:go` returns only `cmd/heygen/builder.go`. Sanitizing at `registerFlag` covers the whole surface.

## Drift-check

Nothing else touched between `670116f` and `9fcfffe2` — delta is exactly the R1-response scope.

## Peer scan / mergeable

- No new reviews after R1; only [`#pullrequestreview-4833096534`](https://github.com/heygen-com/heygen-cli/pull/261#pullrequestreview-4833096534) (mine) on record.
- `pulls/261`: `mergeable: true`, `mergeable_state: "blocked"` (required-review gate, not a code concern).

— *Review by Rames D Jusso*

somanshreddy (commented):
**Review — independent pass at head `9fcfffe`** (not restating Rames's R1; verifying + extending it)

Root cause and fix are correct. `pflag.UnquoteUsage` reads the first back-quoted word in a usage string as the value-type placeholder, so a spec description containing `` `next_token` `` hijacks `--token`'s placeholder. Stripping backticks at `registerFlag` — the single choke point every spec-derived flag passes through — is the right layer, and `registerFlag` is genuinely the only surface that matters here: `UnquoteUsage` only runs on flag *usage* strings, not on cobra `Short`/`Long`, so no other spec-derived text can trigger the placeholder hijack. The "nothing parses the help string" claim holds — `helpText` only feeds the `cmd.Flags().*` usage param; validation reads `flag.Enum` directly, never the rendered help.

**On Rames's R1 (order: sanitize runs on `flag.Help` before the `(required)`/`(allowed:)` suffixes) — resolved, and resolved in the *correct* direction.** I traced the invariant Rames's suggestion (strip the composed string) would have broken:

- Validation is verbatim: `internal/command/spec.go:253` → `slices.Contains(flag.Enum, val)`. The user's input is matched byte-for-byte against `flag.Enum`.
- So if an enum member were `` `fast` ``, the CLI would only accept the literal `` `fast` `` (backticks included). Stripping backticks from the `(allowed: …)` display would then advertise `fast` while the flag rejects `fast` — help that lies about what's accepted.

That's strictly worse than a mangled type placeholder, so **the author's choice to sanitize the description and leave enum values untouched is right, and Rames's "strip the composed string" suggestion should not be adopted.** The explanatory comment + `TestRegisterFlag_SanitizesProseButNotEnumValues` (asserting on the stored `.Usage`, since pflag rewrites backticks at render time) lock the invariant in. Good resolution.

**One non-blocking observation (net-new), for the record — not a change request:**

The comment frames a backticked enum member as "a spec bug worth seeing rather than papering over." In practice the corruption wouldn't read as a spec bug: with a clean description but a backticked enum, the first back-quoted word in the composed usage string now lives in the `(allowed: …)` portion, so pflag would render `--mode fast` as the placeholder *and* strip the backticks from the allowed list at render — i.e. the exact silent corruption this PR fixes for descriptions, just relocated. A user seeing `--mode fast … (allowed: fast, quality)` has no signal that anything is wrong. So the "worth seeing" framing is a touch optimistic.

That said, it's academic: `codegen/grouper.go:585` `formatEnum` builds members from OpenAPI enum arrays via `fmt.Sprintf("%v", …)` — identifier strings, no backticks today — and there's no clean fix anyway (stripping the display would break validation parity as shown above). Leaving enum untouched is the least-bad option. I'd only suggest softening the comment's "worth seeing" wording, or adding a codegen-side assert that enum members are backtick-free so the hypothetical can't silently arrive. Neither blocks.

**Other checks:** the unconditional `strings.ReplaceAll` of the backtick (byte 0x60) is byte-safe on UTF-8 (0x60 never appears as a continuation byte); over-stripping is a non-issue since only prose passes through and a stray/odd backtick removal is strictly an improvement. Tests cover the real end-to-end render path (`HelpStripsBackticksFromSpecDescriptions`) and the enum-preservation invariant. CI green on all three OSes.

**Verdict:** LGTM. R1 addressed correctly; no blocking findings. Only a wording nit on the enum-backtick comment. Not stamping (deferring the approve to Somansh).

— Somu

jrusso1020 (approved):
Approved.

**The enum decision is correct, and I checked it at source rather than taking it on the review discussion.** `validateFlag` in `internal/command/spec.go` compares with `slices.Contains(flag.Enum, val)` — an exact string match, for both the string and string-slice branches. So altering an enum member for display would advertise a value the flag then rejects. Leaving them untouched is right, and the new test pinning that is the correct thing to pin.

**One correction to the inline comment's reasoning, comment-only and non-blocking.** The comment says rewriting an enum value here "would document a value the flag then rejects." That doesn't survive rendering. `pflag.UnquoteUsage` strips the *first* backtick pair from the usage string, not all of them. With the description now sanitized, that first pair is the enum's, so a backticked enum member renders in `(allowed: ...)` without its backticks anyway — and additionally hijacks the type placeholder, so you get `--mode fast` instead of `--mode string`.

The stored usage string keeps them, which is what the new test asserts and why asserting there rather than on `--help` was the right choice. But the rendered help a user actually sees does not. The behavior is still the one I'd want — the enum data is untouched, and the anomaly stays visible rather than being papered over — it's the stated reason that's off by a step. No enum in the spec carries a backtick today, so this is latent either way. Worth a word change so the next reader doesn't trust the premise.

One nit already raised and still open: the commit body argues for sanitizing *after* composition so enum values are covered, which is the opposite of what the code, the inline comment and the test now do. Worth fixing at squash so the history doesn't contradict the change.

Note on process: `/code-review max` can't be invoked programmatically, so I applied its lens by hand rather than running it.

— Rames Jusso

somanshreddy:
Your correction is right, and it goes further than you framed it — it invalidates my *argument*, not just the comment's wording. Fixed at `ae60493` (comment + commit body).

I tested both orderings with a backticked enum member rather than reasoning about it:

| | placeholder | rendered `(allowed: …)` |
|---|---|---|
| description-only (current) | `--mode fast` ✗ | `fast, quality` |
| composed (your R1 suggestion) | `--mode string` ✓ | `fast, quality` |

Identical allowed-lists. `UnquoteUsage` strips the first backtick pair while rendering either way, so "sanitizing enum values would advertise a value the flag rejects" was never a distinction between the two — the rendered help advertises `fast` in **both**, and validation compares against the untouched `flag.Enum` in both. My stated reason was wrong, and so was the reasoning I used to reject your original suggestion.

On rendered output your version is in fact marginally better: same allowed-list, correct placeholder. I've left the code as-is rather than flipping an approved PR over a case that cannot occur — the 169 unique enum values in the live spec contain zero backticks, and structurally they're wire identifiers, not prose. But I want the record straight that this ended up a tie on user-visible behavior with a slight edge to your version, not the clear call I argued for.

The comment now states the boundary as what the CLI is willing to rewrite (prose yes, wire values no) and explicitly says it isn't about rendered output, so the next reader doesn't inherit the bad premise.

Commit body fixed too — it still argued for composed sanitization, which was residue from a `git commit --amend -o` that dropped staged files. Squash-ready now; message, comment, code and test all say the same thing.

PR #366, opened 2026-09-18
update: add --check and stop offering a downgrade to from-source builds (PRINFRA-1331)

somanshreddy:
## Scope

**Surfaces:** CLI | **Module:** Update / release channel

## Summary

`heygen update` on a build compiled from source installed *older* code over the running binary. This
fixes that, and adds `heygen update --check`, which answers whether an update is available without
installing it. Both are prerequisites for a staleness notice, which cannot be correct while the
version gate is wrong.

## Root cause

`validateCurrentVersion` rejected only `""`, `dev`, and `test`. A `go build` from a working tree
stamps a git-describe version:

```
v0.8.1-6-g3ce581a-dirty
```

That parses as valid semver, with prerelease `6-g3ce581a-dirty`. Under semver a prerelease orders
*below* its release, so `v0.8.1-6-g3ce581a-dirty < v0.8.1`, which made `isVersionGreater` true for
the very release the build already contains and more. `update` then "succeeded" by replacing a newer
local build with the last tagged release.

Latent until now because running `update` on a dev build is unusual. It stops being latent the
moment anything checks automatically: a notice would fire forever on developer machines, and opt-in
auto-update would overwrite developer binaries.

## Fix

The gate is now a shape check, not a semver parse, because the string it has to reject parses
perfectly well:

```
^v\d+\.\d+\.\d+(-dev\.[0-9A-Za-z.]+)?$
```

A stable tag or a dev prerelease passes; anything else is a local build.

The dev suffix is deliberately loose about its contents. `dev-release.yml` stamps a timestamp while
the existing test fixtures carry a timestamp plus a sha, so the `dev.` marker is what identifies the
channel. An earlier stricter pattern that pinned the timestamp broke two existing tests, and would
have broken dev-channel updates the next time that format changed.

## How `--check` works

It answers "what would `heygen update` do", without doing it.

The useful distinction is **reports versus refuses**. `update` treats a non-release build and a
package-manager install as terminal errors, because it cannot proceed. `--check` reports both as
fields, so a caller branches on data instead of parsing an error string:

| Field | Meaning |
|---|---|
| `update_available` | a newer release exists upstream |
| `release_build` | false means a local build, so no comparison is possible |
| `install_method` | `direct`, `homebrew`, `npm`, or empty when undetermined |
| `channel` | which release track this build follows |
| `current` / `latest` | the two versions compared |

`update_available` is availability, not installability: a homebrew user gets `true` while `update`
itself would refuse and point at `brew upgrade`. SKILL.md states that distinction, since agents are
the main consumer.

## Design decisions

**`--check` reuses the existing GitHub release source rather than reading the CDN pointer.** One
source means `--check` cannot promise something `update` then contradicts. Reading the CDN was
considered, and matters for a future *passive* check, where a `gh auth token` subprocess per
invocation and GitHub's 60 req/hr/IP limit both bite. Neither applies to a command the user typed.

**Install-method detection failure does not fail the check.** An undeterminable install method should
not block the version comparison the caller actually asked for, so the field is left empty and
documented as meaning undetermined.

**The raw build version is reported, not a canonicalized one.** `canonicalVersion` prepends `v` to
any input, so canonicalizing before validation reported a `dev` build as `vdev`. Canonicalization now
happens only after the shape check passes, and a caller can still see exactly which build it has.

**`--check` with `--version` is a usage error.** Latest-release comparison and exact-version
resolution are different operations. The guard tests flag *presence* rather than a non-empty value,
so `--version ""` is caught too.

**No shared update-plan abstraction yet.** A single read-only planner consumed by both `--check` and
the install path would be cleaner, and is the right move once a passive notice becomes a third
consumer. Extracting it here would mean refactoring the install path inside a change whose purpose is
a downgrade fix, so it is deliberately deferred.

## Testing

35 subtests across the update surface.

Every one was verified by deliberate mutation rather than by passing: reverting the gate, moving
canonicalization back before validation, weakening the flag guard to a value test, hardcoding the
release source to stable, surfacing the install-detection error, and querying the source on a
non-release build each produce a distinct failure, with the tree restored and asserted after each.

Two assertions worth calling out, because they pin things a passing test would otherwise miss. The
table captures the `prerelease` argument handed to the release source and asserts it matches the
reported `channel`, so reporting `dev` while querying stable fails. And both error paths unmarshal
stderr and assert `error.code`, not just the message text.


jrusso1020 (approved):
**Approving.** Reviewed at `e74d7f7a64c21f35b98d159bf3e1ece79b205d23`. The gate is correct over the complete set of version strings this repo's three producers can emit — I checked that by running the classifier against every tag the repo has ever published rather than against a sample. One finding, non-blocking, in the place you pointed at.

## Verified by measurement, not by reading

**The classifier against the real release history.** I pulled all 138 published tags and ran each through `validateCurrentVersion`:

| | count | accepted | channel |
|---|---|---|---|
| dev prereleases | 109 | 109 | `dev` |
| stable tags | 29 | 29 | `stable` |
| anything else | 0 | — | — |

**0 rejected.** No published build loses the ability to update.

**The reject side, enumerated over what the producers actually emit.** `Makefile:1` is `git describe --tags --always --dirty`; `.goreleaser.yaml:18` stamps `{{.Tag}}`; `main.go:19` defaults to `dev`. Every shape those three can produce:

| input | verdict | where it comes from |
|---|---|---|
| `v0.8.1` | accept, `stable` | goreleaser, or describe exactly on a clean stable tag |
| `0.8.1` | accept, `stable` | ldflags override without the `v` |
| `v0.8.2-dev.202609181200` | accept, `dev` | goreleaser on a dev tag |
| `v0.0.1-dev.20260407.c9e2db1` | accept, `dev` | goreleaser, historical dev format |
| `v0.8.1-dirty` | reject | describe, clean tag + dirty tree |
| `v0.8.1-6-g3ce581a` | reject | describe, past a stable tag |
| `v0.8.1-6-g3ce581a-dirty` | reject | the bug's own case |
| `v0.8.2-dev.202609181200-dirty` | reject | describe, on a dev tag, dirty |
| `v0.8.2-dev.202609181200-6-gabc1234` | reject | describe, past a dev tag |
| `v0.0.1-dev.20260407.c9e2db1-4-gdeadbee` | reject | describe, past a date+sha dev tag |
| `3ce581a` | reject | `--always`, repo with no tags |
| `dev` | reject | plain `go build` |
| `v1.0.0-rc.1` | reject | not a channel this repo has |
| `v0.8.1+build.5` | reject | build metadata |
| `v0.8.1-dev.` / `-dev.1..2` / `-dev.1.` | reject | regex passes, `semver.NewVersion` catches it |
| `v01.8.1` | reject | regex passes, semver rejects the leading zero |
| `v0.8.1-DEV.1` | reject | `-dev.` is case-sensitive; the producer is lowercase |

The two-stage check earns its keep on three of those: `-dev.1..2`, `-dev.1.` and `v01.8.1` all satisfy the shape and are stopped by the semver parse.

**Your root cause, checked at source rather than taken from the description.** `semver.NewVersion` orders `v0.8.1-6-g3ce581a-dirty` **below** `v0.8.1`, and `isVersionGreater("v0.8.1", "v0.8.1-6-g3ce581a-dirty")` returns `true`. The downgrade was real.

**The downgrade path now refuses out loud.** Built a binary at `-X main.version=v0.8.1-6-g3ce581a-dirty` and ran it:

```
$ heygen update
{"error":{"code":"error","message":"current build version is not release-tagged; reinstall from a release build to use heygen update"}}
exit=1
```

Not a silent no-op, and the message names the remedy. The other no-op path — release build, nothing newer upstream — still reports `heygen is already at <v>` at exit 0 via `rel = updateRelease{Version: current}` (`update.go:247`). Both states are legible.

**`--check` end to end, against the live release source, no API key set:**

```
$ heygen update --check                       # stamped v0.1.0
{"current":"v0.1.0","latest":"v0.8.1","update_available":true,"release_build":true,
 "install_method":"direct","channel":"stable","message":"heygen v0.8.1 is available; you have v0.1.0"}
exit=0

$ heygen update --check                       # stamped v0.8.1-6-g3ce581a-dirty
{"current":"v0.8.1-6-g3ce581a-dirty","latest":"","update_available":false,"release_build":false,
 "install_method":"direct","channel":"","message":"not a release build, so no update can be offered"}
exit=0

$ heygen update --check --version v0.2.0
{"error":{"code":"usage_error","message":"--check reports what an update would do; it cannot be combined with --version"}}
exit=2
```

`--human` renders both correctly. "35 subtests across the update surface" is exact — 18 top-level plus 17 subtests. `go vet` clean, full suite green, both changed Go files gofmt-clean (the repo's 4 pre-existing unformatted files are untouched).

## Your "two existing tests" — a design, and better founded than you argued it

You flagged the loose dev suffix as the part you were least sure of, and framed the justification as *the existing test fixtures carry a timestamp plus a sha*. Fixtures are a weak warrant, so I went looking for the producer. The format is real:

`dev-release.yml` at `b8981b72` (*scripts: rewrite installer for versioned releases (#49)*) computed

```
version="v${base}-dev.$(date -u +%Y%m%d).${short_sha}"
```

and **three tags in that format are published today**: `v0.0.1-dev.20260407.c9e2db1`, `v0.0.1-dev.20260407.47546c3`, `v0.0.1-dev.20260407.465046f`. Every other dev tag — the other 106 — uses the current 12-digit `YYYYMMDDHHmm`.

So the answer to "do those tests encode a contract or the old bug" is **contract**. A pattern pinning the timestamp to 12 digits would strand three real builds. The looseness is right, and the two tests are right to defend it.

The one thing to change is where the comment points. `update.go:39-42` cites the fixtures; the actual warrant is the tag history. Worth saying so, because a fixture reads as something you can update and a published release does not.

## Finding — the exclusion of `-` from the char class is load-bearing, and nothing pins it

`[0-9A-Za-z.]+` is what rejects a from-source build that descends from a **dev** tag, since `git describe` appends `-<N>-g<sha>` and `-dirty`. That is the whole reason `v0.8.2-dev.202609181200-6-gabc1234` fails the gate. The comment above the regex explains why the suffix is loose but never says which character keeps it sound.

Measured with a mutation battery — 9 mutations, 8 caught, 1 survived:

| mutation | result |
|---|---|
| **`[0-9A-Za-z.]+` → `[0-9A-Za-z.-]+`** | **survives the entire repo suite** |
| revert the gate to `raw == "" \|\| "dev" \|\| "test"` | caught (8 failures) |
| unanchor the regex (drop `^`/`$`) | caught (8) |
| shape-check `raw` instead of the canonicalized string | caught (4) — `stable tag without v` |
| `updateChannel` always returns `stable` | caught (2) — the `prerelease`-vs-`channel` assertion |
| canonicalize the reported `Current` up front | caught (2) — `literal dev build keeps its raw version` |
| `--check`/`--version` guard `Changed()` → value test | caught (1) |
| surface the install-detection error | caught (1) |
| query the release source on a non-release build | caught (1) |

Under the surviving mutation, all four describe-off-a-dev-tag shapes become "release builds", route to the dev channel, and compare like this:

```
v0.8.2-dev.202609181200-6-gabc1234         ACCEPT ch=dev  isVersionGreater("v0.8.2", it) = true
v0.8.2-dev.202609181200-6-gabc1234-dirty   ACCEPT ch=dev  isVersionGreater("v0.8.2", it) = true
v0.8.2-dev.202609181200-dirty              ACCEPT ch=dev  isVersionGreater("v0.8.2", it) = true
v0.0.1-dev.20260407.c9e2db1-4-gdeadbee     ACCEPT ch=dev  isVersionGreater("v0.8.2", it) = true
```

That is this PR's bug, reinstated for anyone building from source while the newest reachable tag happens to be a dev tag.

Why the table misses it: of its 10 cases, 2 exercise the dev suffix and both are accept-side. All four reject cases fail on a *different* part of the pattern — they contain no `-dev.` at all, so they never reach the char class. The suffix has an accept boundary and no reject boundary.

One row closes it, and it is the row the comment should be naming:

```go
{"git-describe off a dev tag", "v0.8.2-dev.202609181200-6-gabc1234", false},
```

Non-blocking — the code is correct as written. This is about keeping it correct through the next edit to that regex, which the current comment invites by explaining only the looseness.

## Smaller notes

**`--version` still installs a downgrade without comment.** `update.go:234-262` resolves the target and installs it whenever `rel.Version != current`, with no `isVersionGreater` check — `--version v0.1.0` from `v0.8.1` reports `Updated heygen to v0.1.0`. Pre-existing, and defensible as an explicit pin, but worth knowing it survives a change whose title is about not offering downgrades. `TestUpdate_...` at `update_test.go:252` locks that behavior in, so if it is intended, it already is.

**The canonicalization sentence in the description is inverted.** *"Canonicalization now happens only after the shape check passes"* — in `validateCurrentVersion:276` it happens **before**, and that ordering is what makes `0.8.1` accepted at `update_test.go:284`. The claim is true of the reported `Current` field, which is what the bullet is really about. Flagging only so a later reader doesn't "fix" the order and drop the no-`v` case.

**RELEASE.md is now behind the code in two places.** The format table at `RELEASE.md:158-162` lists stable, dev-with-12-digit-timestamp, and `dev` — not the date+sha dev shape that three published tags use, which is exactly the shape justifying the loose regex. And *Update channels* (`RELEASE.md:191-198`) documents two channels and neither `--check` nor the new third state where `update` refuses. README and SKILL.md both got updated; RELEASE.md is the canonical reference for this surface and didn't.

**`message` is a seventh stdout field** not in SKILL.md's list. Harmless, and arguably right to leave prose out of an agent contract, but a caller unmarshalling into a strict struct will meet it.

**Cosmetic:** on the non-release path `--human` prints `Channel:` and `Latest:` as empty rows. Consistent with how the generic formatter treats empty strings elsewhere, so not something to fix here.

Good separation of *reports* from *refuses*, and `TestUpdateCheck_Reports` asserting the `prerelease` argument actually handed to the release source against the reported `channel` is the assertion that makes the channel field trustworthy rather than decorative.

_Review by Rames_

